PR #4, opened 2016-03-11
Override defaultResolverFn to use PropertyAccess component

### LINKED ISSUE #2: Bundle philosophy
I'm trying to understand core concepts of this bundle.
Could you please validate or invalidate my knowledge about it please ?

First, overblog/GraphQLBundle does not use directly persistence layer (or doctrine service - different from https://github.com/4rthem/graphql-mapper (-bundle) that it seems hardly couple with Doctrine). It uses services as resolver to fetch data (from database using doctrine or not, in memory, etc.).
In code examples, where it is called `$humans = StarWarsData::humans();`, I had to replace this by my entity repository for exemple.
In resolveFriend methods, I could replace `StarWarsData::getFriends($character);` by `$character->getFriends()` in traditional Symfony/Entity/Doctrine way.

As soon as i have a property with a listOf personal type, defining a object, i have to set a resolver, to tell bundle how to retreive data.

If I would like to expose the full list of my Characters, i declare the following Query object:

```yml
# This implements the following type system shorthand:
#   type Query {
#     characters(limit: Int, offset: Int, orderBy: String): [Character]
#   }
#
Query:
    type: object
    config:
        description: "Full list of characters in the Star Wars universe."
        fields:
            characters:
                type: "[Character]"
                args:
                    limit:
                        description: "If omitted,return 10 characters maximum"
                        type: "Int"
                    offset:
                        description: "If omitted, begin at 0"
                        type: "Int"
                    orderBy:
                        description: "If omitted, no sort applied"
                        type: "String"
                resolve: "@=service('acme_bundle.my_service').find([args])"
```
I set a different resolve expression to be sure to understand the way to use it.

Thanks for your help

mcg-web:
Should fix #2 but the true place of defaultResolverFn is in php lib [webonyx/graphql-php](https://github.com/webonyx/graphql-php).

### DIFF
--- a/Resolver/ConfigResolver.php
+++ b/Resolver/ConfigResolver.php
@@ -11,6 +11,7 @@
 
 namespace Overblog\GraphQLBundle\Resolver;
 
+use GraphQL\Executor\Executor;
 use GraphQL\Type\Definition\ResolveInfo;
 use Overblog\GraphQLBundle\Definition\Argument;
 use Overblog\GraphQLBundle\Definition\ArgsInterface;
@@ -44,6 +45,11 @@ class ConfigResolver implements ResolverInterface
     private $argResolver;
 
     /**
+     * @var callable
+     */
+    private $defaultResolveFn = ['GraphQL\Executor\Executor', 'defaultResolveFn'];
+
+    /**
      * @var array
      *            [name => callable]
      */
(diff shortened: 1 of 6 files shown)


mcg-web:
related PR https://github.com/webonyx/graphql-php/pull/28

Issue #17, opened 2016-03-24
Add a batching layer


ooflorent:
When using nested routes in Relay, multiple queries are required to render the application. Relay default network layer receives the requests as an array but runs then concurrently. Because of query structure, it is really hard to combine them in a single query (fragment names and variables can create conflicts).

An alternate solution is to support query batching. Instead of receiving a json request containing the query and the variables, the server should be able to decode an array of requests.

```js
// Normal query
{query: "query q{user(id:$id){id}}", variables: {id: 1}}

// Batch alternative
[
  {query: "query q{user(id:$id){id}}", variables: {id: 1}},
  {query: "query q{user(id:$id){id}}", variables: {id: 2}},
]

// Combine approach (too hard to implement)
{query: "query q{u0:user(id:$id0){id}u1:user(id:$id1){id}}", variables: {id0: 1, id1: 2}}
```

PR #18, opened 2016-03-24
Add a batching layer

mcg-web:
Add query parameter `batch` on graphQL endpoint to enabled batching (see #17).


Controller/GraphController.php
```diff
@@ -19,9 +19,32 @@ class GraphController extends Controller
...
+            $data = $this->treatBatchQuery($request);
+        } else {
+            $data = $this->treatNormalQuery($request);
+        }
 
-        return new JsonResponse($res->toArray(), 200);
+        return new JsonResponse($data, 200);
+    }
+
+    private function treatBatchQuery(Request $request)
```
ooflorent: Can you replace `treat` by `process`?

Issue #45, opened 2016-05-30
Class 'Overblog\GraphQLBundle\__DEFINITIONS__\RangeFilterType' not found


VincentClair:
Hello,

I'm trying to make a query on products with a custom filter type.
Here is the base `Query.types.yml`:
```yaml
Query:
    type: object
    config:
        description: "Product bundle."
        fields:
            getProducts:
                type: "QueryProduct"
                args:
                    filter:
                        description: "Product search filters."
                        type: "ProductFilter"
 ```

My `ProductFilter.types.yml` is defined as an `input-object`:
```yaml
ProductFilter:
    type: input-object
    config:
        description: "Product filters"
        fields:
            categoryIds:
                type: "[Int]"
                description: "Category filter."
            highLight:
                type: "ProductHighLightFilter"
                description: "Highlight product filter."
            ...
            price:
                type: "RangeFilter"
                description: "Product price filter."
            ...
```

 And finally my `RangeFilter.types.yml`, as an `input-object` too:
```yaml
RangeFilter:
    type: input-object
    config:
        description: "Range filter"
        fields:
            min:
                type: "Float"
                description: "Min value."
            max:
                type: "Float"
                description: "Max value."
```

When I run the tests (with Codeception) just after initialize the app (clear cache, install database and load fixtures, create admin user), i get randomly the following error `Class 'Overblog\GraphQLBundle\__DEFINITIONS__\RangeFilterType' not found`.
I get this only for RangerFilter, that is an "inline-object" too.

Have you an idea where the error come from ? Am i using the "inline-object" correctly ?

mcg-web:
Can you try to run your tests in isolated process please (like phpunit)?

VincentClair:
Thanks for your reply. I ran only the crashed test and here is the result:
```
Api Tests (3) -------------------------------------------------------------------------------------------------
Modules: Symfony2, Doctrine2, REST, \Helper\Api
---------------------------------------------------------------------------------------------------------------
Get all manufacturers (Loyalty\ProductBundle\Tests\GraphQL\GraphQLCest::getManufacturers)
Scenario:
* I send post "/auth/login",{"_username":"admin","_password":"admin"}

  [Request headers] []
  [Request] POST /auth/login {"_username":"admin","_password":"admin"}
  [Response] 
  [Headers] {"cache-control":["no-cache"],"x-debug-token":["0d95df"],"Set-Cookie":[{},{}]}
  [Status] 204
* I see response code is 204
[...]

PR #46, opened 2016-05-31
Type class loader bugfix and more

mcg-web:
* Fix type class loader
* Document Union and InputObject definition
* Travis required Symfony 3.1 for test


mcg-web:
should fix #45

PR #64, opened 2016-09-20
Exceptions can be mapped to UserError and UserWarning

adri:
First of all, this is a great library! I really like using it 👍 

### Context
When writing resolvers for mutations or queries I often convert exceptions from the business layer into errors or warnings visible to the GraphQL API user. A mutation resolver can look like this:

```php
    public function someMutation($someId)
    {
        try {
            // do something
        } catch (SomethingWentWrong $e) {
            throw new UserError($e->getMessage());
        } catch (AnotherError $e) {
            throw new UserError($e->getMessage());
        } catch (ShouldNotHappen $e) {
            throw new UserError($e->getMessage());
        } catch (\InvalidArgumentException $e) {
            throw new UserError($e->getMessage());
        }
    }
```

Doing this manually for all mutations adds some bloat. This is the same for REST APIs which is why FOSRestBundle added the [`exception` configuration](http://symfony.com/doc/current/bundles/FOSRestBundle/4-exception-controller-support.html).

### What has been done
- Added a configuration to map exceptions that should be displayed as warnings or errors
- Added unit tests for the configuration
- Added a functional test
- Wrote documentation in the readme how to use it

### How to test
- For an example how to use it, please see the updated readme

### Notes
- I had to use `@runInSeparateProcess` on the Extension test-case because of this [`static $config`](https://github.com/overblog/GraphQLBundle/blob/master/DependencyInjection/OverblogGraphQLExtension.php#L76). The total test runtime isn't hit too much by it.


Error/ErrorHandler.php
```diff
@@ -129,4 +133,32 @@ public function handleErrors(ExecutionResult $executionResult, $throwRawExceptio
...
+     * @param \Exception $rawException
+     * @return \Exception
+     */
+    protected function convertException(\Exception $rawException = null)
+    {
+        if (empty($rawException)) {
+            return $rawException;
+        }
+
+        $types = [
```
ooflorent: I think we should make this configurable via service parameters.

Error/ErrorHandler.php
```diff
@@ -129,4 +133,32 @@ public function handleErrors(ExecutionResult $executionResult, $throwRawExceptio
...
+    /**
+     * Tries to convert a raw exception into a user warning or error
+     * that is displayed to the user.
+     *
+     * @param \Exception $rawException
+     * @return \Exception
+     */
+    protected function convertException(\Exception $rawException = null)
+    {
+        if (empty($rawException)) {
```
ooflorent: This test could be rewritten to make it easier to understand. Maybe something like:

```php
if (null === $rawException) {
    return null;
}
```

Error/ErrorHandler.php
```diff
@@ -26,13 +26,17 @@ class ErrorHandler
...
+    private $exceptionMap;
+
+    public function __construct($internalErrorMessage = null, LoggerInterface $logger = null, array $exceptionMap = [])
     {
         $this->logger = (null === $logger) ? new NullLogger() : $logger;
         if (empty($internalErrorMessage)) {
             $internalErrorMessage = self::DEFAULT_ERROR_MESSAGE;
         }
         $this->internalErrorMessage = $internalErrorMessage;
+        $this->exceptionMap = $exceptionMap;
```
ooflorent: `exceptionMap` could be transformed to improve the lookup. Currently `convertException` must loop over every entries to find the error class. If we flatten it we could greatly improve the lookup.

```php
// e.g.
[
  'Symfony\Component\Routing\Exception\ResourceNotFoundException' => 'warning',
  'InvalidArgumentException' => 'error',
]

**Comment by ooflorent** [comment]
Thank you! Your contribution is amazing! I'm nitpicking but this PR could be merged in its current state. cc @mcg-web

File: Error/ErrorHandler.php (hunk 1 above)
Comment (mcg-web) [inline Error/ErrorHandler.php]: Same here, you can added to the service arg [here](https://github.com/adri/GraphQLBundle/blob/4308b8c3cd1a367a7ffdda56f3fddfba222877a9/DependencyInjection/OverblogGraphQLExtension.php#L109)...

File: Error/ErrorHandler.php [hunk 4]
```diff
@@ -129,4 +133,32 @@ public function handleErrors(ExecutionResult $executionResult, $throwRawExceptio
...
+            return $rawException;
+        }
+
+        $types = [
+            'warnings' => 'Overblog\\GraphQLBundle\\Error\\UserWarning',
+            'errors' => 'Overblog\\GraphQLBundle\\Error\\UserError',
+        ];
+
+        foreach ($types as $type => $errorClass) {
+            if (!empty($this->exceptionMap[$type])
```
mcg-web: After flipping the exception map this will be simplify.

Error/ErrorHandler.php
adri: Thanks for the review!

What do you think of mapping exceptions directly to error classes? Then we don't have to add another parameter to the class and the lookup is even quicker. Memory consumption might be a bit higher.


```php
[
    'Symfony\Component\Routing\Exception\ResourceNotFoundException' => 'Overblog\\GraphQLBundle\\Error\\UserWarning',
    'InvalidArgumentException' => 'Overblog\\GraphQLBundle\\Error\\UserError',
]
```

adri:
I changed the exception map to map from `custom exception => error class` to speed up the exception matching. Building the mapping is done on container compile time to use cache power.

The type mapping (`warnings => UserWarning` class) is done in configuration and can be overwritten.

Error/ErrorHandler.php
ooflorent: Yes it could be done this way. But if these classes are set using a setter then `ErrorHandler` would have to reprocess `exceptionMap`.

mcg-web:
Thank you @adri

adri:
Thanks for reviewing and merging so quickly!

adri:
Want to do a release? :D

mcg-web:
Yes we can release it in a new version.

mcg-web:
[v0.6.0](https://github.com/overblog/GraphQLBundle/releases/tag/v0.6.0) released...

adri:
That was quick! thanks man! :)

PR #73, opened 2016-11-05
Promise

mcg-web:
Enable promise support


Resolver/AccessResolver.php
```diff
@@ -38,33 +39,41 @@ private function filterResultUsingAccess(callable $accessChecker, callable $reso
...
-                    },
-                    $result
-                );
-                break;
-            case $result instanceof Connection:
-                $result->edges = array_map(
-                    function (Edge $edge) use ($accessChecker, $resolveArgs) {
-                        $edge->node = $this->hasAccess($accessChecker, $edge->node, $resolveArgs) ? $edge->node : null;
+        $onFulFilled = function ($result) use ($accessChecker, $resolveArgs) {
+            switch (true) {
```
ooflorent: `if/else` blocks would be easier to read

adri:
Looks cool! Does this enable a [Data Loader](https://github.com/facebook/dataloader) style library to batch/unique SQL queries?

Are there otherwise currently options how to minimalise queries?

vladar:
@adri Recently added docs on how to do this in stock graphql-php: http://webonyx.github.io/graphql-php/data-fetching/#solving-n1-problem (will be available since next version 0.9)

mcg-web:
@adri yes the job has be done in the GraphQL lib now missing the bunble implementation. A big feature coming soon as possible, but missing some work to be complete:

* Implement the deferred behavior in our [dataloader](https://github.com/overblog/dataloader-php) (wip [here](https://github.com/mcg-web/dataloader-php/tree/webonyx-graphql-sync-promise))
* Make it easy to use dataloader with the bundle
* And a lot of tests to validate all these.

adri:
Awesome, thanks for the information!

vladar:
@mcg-web By the way, if you just add Generic `thenable` support in your DataLoader - it should be enough. `GraphQL\Deferred` has `then` method - https://github.com/webonyx/graphql-php/blob/master/src/Deferred.php#L45-L48 which returns valid A+ promise.

So I assumed it could be as simple as:

```php
$loader = new DataLoader(function($keys) {
    return new GraphQL\Deferred(function() use ($keys) {
       if ($ok) {
           return $db->loadKeys($keys);
       } else {
          throw new \Exception("Some error"); // shoud properly translate to $onRejected callback in `then` method
       }
    });
}));
```

But I didn't look into the source code, so I may be missing something.

mcg-web:
@vladar the difficulty I had with the native deferred supports, it's that it's not the promise that's return. I need to trigger the DataLoader::await() to resolve the promise. I'll try to take back from the start to validate that.

vladar:
@mcg-web Technically we implemented a kind of event-loop in graphql-php - [here](https://github.com/webonyx/graphql-php/blob/master/src/Executor/Promise/Adapter/SyncPromiseAdapter.php#L131-L137).

So you simply don't need to call `DataLoader::await()` for promises/deferreds of graphql-php (same as you don't need to call it for any event-loop-based environment). Executor will run those deferreds itself [here](https://github.com/webonyx/graphql-php/blob/master/src/Executor/Executor.php#L116) (which will in turn resolve/reject).

But yeah, I may be missing something. Anyway, send me a note if you need to clarify something with those sync promises/Deferred implementation.

Resolver/AccessResolver.php
Nyalab: i agree, please use `if` statements

Resolver/AccessResolver.php
```diff
@@ -38,33 +39,41 @@ private function filterResultUsingAccess(callable $accessChecker, callable $reso
...
-                }
-                break;
+                            return $edge;
+                        },
+                        $result->edges
+                    );
+                    break;
+                default:
+                    if (!$this->hasAccess($accessChecker, $result, $resolveArgs)) {
+                        throw new UserWarning('Access denied to this field.');
```
Nyalab: you can't expose which field is it?

adri:
Is there help needed with tests or anything else for the promise integration?

mcg-web:
Thanks @adri, but now that the v0.9 was released, we should first fix some compatibility issues, before adding promise. I'm working on this...

Executor/Promise/Adapter/ReactPromiseAdapter.php
```diff
@@ -0,0 +1,77 @@
...
+     */
+    public function wait(Promise $promise)
+    {
+        if (!$this->isThenable($promise)) {
+            return;
+        }
+        $wait = true;
+        $resolvedValue = null;
+        $exception = null;
+        /** @var \React\Promise\PromiseInterface $reactPromise */
```
ooflorent: Remove this please

Resolver/AccessResolver.php
```diff
@@ -38,30 +48,39 @@ private function filterResultUsingAccess(callable $accessChecker, callable $reso
...
+            $result->edges = array_map(
+                function (Edge $edge) use ($accessChecker, $resolveArgs) {
+                    $edge->node = $this->hasAccess($accessChecker, $edge->node, $resolveArgs) ? $edge->node : null;
+
+                    return $edge;
+                },
+                $result->edges
+            );
+        } elseif (!$this->hasAccess($accessChecker, $result, $resolveArgs)) {
+            throw new UserWarning('Access denied to this field.');
```
ooflorent: Please add the field name into the error message

mcg-web: the name can be found using the path, here an example:

```json
{
  "data": {
    "user": {
      "name": null
    }
  },
  "extensions": {
    "warnings": [
      {
        "message": "Access denied to this field.",
        "locations": [
          {
            "line": 1,
            "column": 24
          }
        ],
        "path": [
          "user",
          "name"
        ]
      }
    ]
  }
}
```

adri:
@OwlyCode Thanks a lot for your example `DataLoaderAdapter`. Would it be handy to include this in this project?  Just like `overblog_graphql.react.promise_adapter` but then `overblog_graphql.data_loader_react.promise_adapter`?

mcg-web:
Hi @adri! you can use dataloader with this project with the lib directly [here an example](https://github.com/mcg-web/sandbox-dataloader-graphql-php) in [this file](https://github.com/mcg-web/sandbox-dataloader-graphql-php/blob/master/app/with-dataloader-native-promise.php)... Or the full symfony way using the dataloader bunfle with this config:

```yaml
#graphql
overblog_graphql:
    definitions:
        schema:
            query: Query
    services:
        promise_adapter: "webonyx_graphql.sync_promise_adapter"

#dataloader
overblog_dataloader:
    defaults:
        promise_adapter: "overblog_dataloader.webonyx_graphql_sync_promise_adapter"
    loaders:
        ships:
            alias: "ships_loader"
            batch_load_fn: "@app.graph.ships_loader:all"
```

Batch loader function:

```yaml
services:
    app.graph.ship_repository:
        class: AppBundle\Entity\Repository\ShipRepository
        factory: ["@doctrine.orm.entity_manager", getRepository]
        arguments:
            - AppBundle\Entity\Ship

    app.graph.ships_loader:
        class: AppBundle\GraphQL\Loader\ShipLoader
        arguments:
            - "@overblog_graphql.promise_adapter"
            - "@app.graph.ship_repository"
```

```php
<?php

namespace AppBundle\GraphQL\Loader;

use AppBundle\Entity\Repository\ShipRepository;
use GraphQL\Executor\Promise\PromiseAdapter;

class ShipLoader
{
    private $promiseAdapter;

    private $repository;

    public function __construct(PromiseAdapter $promiseAdapter, ShipRepository $repository)
    {
        $this->promiseAdapter = $promiseAdapter;
        $this->repository = $repository;
    }

    public function all(array $shipsIDs)
    {
        $qb = $this->repository->createQueryBuilder('s');
        $qb->add('where', $qb->expr()->in('s.id', ':ids'));
        $qb->setParameter('ids', $shipsIDs);
        $ships = $qb->getQuery()->getResult();

        return $this->promiseAdapter->all($ships);
    }
}
```

Using it in a resolver:

```php
    public function resolveShip($shipsID)
    {
        $promise = $this->container->get('ships_loader')->load($shipsID);

        return $promise;
    }
```

this an example using the sync promise adapter of Webonyx.

I'm working on a full example but missing some time at the moment...

adri:
Thanks @mcg-web :)
I tested this configuration and it works great!

But when using `react/promise` I got an infinite loop in the `ReactPromiseAdapter::wait` code. When using the implementation above from @OwlyCode which is calling `DataLoader::await` it works just fine :)

PR #75, opened 2016-11-10
fix(README): StarWarsData namespace doc was wrong

remy-theroux:
Fix namespace for StarWarsData in README example


mcg-web:
hi and thanks for the fix :+1:

PR #76, opened 2016-11-17
Avoid creating services before compiler passes are all done

alafon:
Issue `$container->get()` should not be called during compiler passes to avoid issues where a resolver service (for instance), implements ContainerAwareInterface and for which a service is not fully yet created.

This fix uses `$container->findDefinition()` instead and checks if the class (string) is a sub class of the interface.

### DIFF
--- a/DependencyInjection/Compiler/TaggedServiceMappingPass.php
+++ b/DependencyInjection/Compiler/TaggedServiceMappingPass.php
@@ -12,7 +12,6 @@
 namespace Overblog\GraphQLBundle\DependencyInjection\Compiler;
 
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
-use Symfony\Component\DependencyInjection\ContainerAwareInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Reference;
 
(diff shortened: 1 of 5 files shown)


DependencyInjection/Compiler/TaggedServiceMappingPass.php
```diff
@@ -43,9 +42,9 @@ public function process(ContainerBuilder $container)
         foreach ($mapping as $name => $options) {
             $cleanOptions = $options;
             $solutionID = $options['id'];
-            $solution = $container->get($solutionID);
 
-            if ($solution instanceof ContainerAwareInterface) {
+            $definition = $container->findDefinition($solutionID);
+            if (is_subclass_of($definition->getClass(), 'Symfony\Component\DependencyInjection\ContainerAwareInterface')) {
```
ooflorent: I don't like this. Using the `::class` keywords would be better.

```php
if (is_subclass_of($definition->getClass(), ContainerAwareInterface::class)) {
```

mcg-web: ::class not compatible with php 5.4...

Tests/DependencyInjection/Compiler/ResolverTaggedServiceMappingPassTest.php
```diff
@@ -0,0 +1,96 @@
...
+
+    public function testCompilationWorksPassConfigDirective()
+    {
+        $this->addCompilerPassesAndCompile();
+
+        $this->assertTrue($this->container->has('fake_resolver'));
+    }
+}
+
+class FakeCompilerPass implements CompilerPassInterface
```
mcg-web: Create separate files for each class to respect  psr4 please, thanks for the work :)

PR #78, opened 2016-11-22
Added paginator helper to convert first/after and before/last to offset/limit

OwlyCode:
This adds a standalone object to help on pagination. Its role is to convert GraphQL arguments from the ConnectionArgs to an offset/limit pair. For instance : 

```php
    public function resolveList($args)
    {
        $pagination = new EntityPaginator(function ($offset, $limit) {
            return $this->myRepository->getAll($offset, $limit); // Or Doctrine or whatever
        });
        
        // Make use of the 'first' and 'after' arguments
        return $pagination->forward($args);
    }
```

All you have to do is provide a callable that fetch the items you need to paginate.

For backward pagination, you have to provide an item count : 

```php
return $pagination->backward($args, $this->myRepository->count());
```

You can also have an automatic mode that detects if you go backward or forward : 

```php
return $pagination->auto($args, $this->myRepository->count());
```

To avoid computing a total that is uneeded for forward pagination, you can provide a callable that will only be resolved when going backward : 

```php
return $pagination->auto($args, [$this->myRepository, 'count']);
```

### DIFF
--- a/Relay/Connection/Paginator.php
+++ b/Relay/Connection/Paginator.php
@@ -1,0 +1,90 @@
+<?php
+
+namespace Overblog\GraphQLBundle\Relay\Connection;
+
+use Overblog\GraphQLBundle\Definition\Argument;
+use Overblog\GraphQLBundle\Relay\Connection\Output\Connection;
+use Overblog\GraphQLBundle\Relay\Connection\Output\ConnectionBuilder;
+
+class Paginator
+{
+    /**
+     * @var callable
+     */
+    private $fetcher;
+
+    /**
+     * @param callable $fetcher
+     */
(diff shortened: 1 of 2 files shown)


Nyalab:
👏

ooflorent:
Damn! This is good.

Relay/Connection/Paginator.php
```diff
@@ -0,0 +1,90 @@
...
+            return $this->forward($args);
+        }
+    }
+
+    /**
+     * @param Argument|array $args
+     *
+     * @return Argument
+     */
+    private function protectArgs($args)
```
ooflorent: `toArgument` would be a better name

PR #88, opened 2016-12-20
Paginator backward method now accepts arguments for its callback

alafon:
Make it possible to pass arguments for the callable used in the `auto` and `backward` function of the `Paginator`.

### DIFF
--- a/Relay/Connection/Paginator.php
+++ b/Relay/Connection/Paginator.php
@@ -33,11 +33,14 @@ class Paginator
     /**
      * @param Argument|array $args
      * @param int|callable   $total
+     * @param array          $callableArgs
      *
      * @return Connection
      */
-    public function backward($args, $total)
+    public function backward($args, $total, array $callableArgs = [])
     {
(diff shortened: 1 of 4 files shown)


mcg-web:
hi @alafon , with this change, is the documentation up to date?

alafon:
@mcg-web actually there was no documentation about the `auto()` or `backward()` so no update is needed. But I'll improve it.

mcg-web:
ok nice, thank you :+1:

mcg-web:
tips: use cli `php-cs-fixer fix .` or add the header to fix ci build...

alafon:
I did but like this `vendor/bin/php-cs-fixer fix --dry-run --diff Tests/Relay/Connection/`...
Didn't know that php-cs-fixer does not use the configuration file when using a target... now I know.

alafon:
Added a small piece of doc with a "silly" example, but it does the job :)

mcg-web:
thank you @alafon for contribution ;)

PR #97, opened 2017-01-02
Builders

mcg-web:
| Q             | A
| ------------- | ---
| Bug fix?      | No
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| License       | MIT

- Allow Args builder config
- Better field builder documentation

### DIFF
--- a/Resources/doc/definitions/builders/args.md
+++ b/Resources/doc/definitions/builders/args.md
@@ -1,4 +1,91 @@
 Args builder
 ============
 
-TODO
+Builder is a way to don't repeat args definition.
+
+Define your custom args builder
+
+```yaml
+#app/config/config.yml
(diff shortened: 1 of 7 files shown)


Resources/doc/definitions/builders/args.md
```diff
@@ -1,4 +1,91 @@
...
+
+```yaml
+#app/config/config.yml
+overblog_graphql:
+    #...
+    definitions:
+        #...
+        builders:
+            args:
+                -
```
Nyalab: is this normal?

mcg-web: it means that we need a list of alias/class values

PR #100, opened 2017-01-12
Use the whatwg-fetch polyfill for GraphiQL

pascaldevink:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | N/A
| Fixed tickets | none
| License       | MIT

When using GraphiQL with browsers that do not natively support `window.fetch` (such as Safari in my case, but also older versions of Chrome, Firefox and IE), javascript errors pop up and GraphiQL becomes unusable.

screenshot 2017-01-12 11 13 58

The example HTML that GraphiQL themselves provide uses the whatwg-fetch polyfill as a workaround.

See https://github.com/graphql/graphiql/blob/master/example/index.html#L31 and https://www.npmjs.com/package/whatwg-fetch for more information.

### DIFF
--- a/Resources/views/GraphiQL/index.html.twig
+++ b/Resources/views/GraphiQL/index.html.twig
@@ -13,6 +13,7 @@
   </style>
   <link href="https://unpkg.com/graphiql@{{ versions.graphiql }}/graphiql.css" rel="stylesheet">
   {% endblock style %}
+  <script src="https://unpkg.com/whatwg-fetch@{{ versions.fetch }}/fetch.js"></script>
   <script src="https://unpkg.com/react@{{ versions.react }}/dist/react.min.js"></script>
   <script src="https://unpkg.com/react-dom@{{ versions.react }}/dist/react-dom.min.js"></script>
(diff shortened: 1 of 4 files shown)


Resources/views/GraphiQL/index.html.twig
```diff
@@ -13,6 +13,7 @@
   </style>
   <link href="https://unpkg.com/graphiql@{{ versions.graphiql }}/graphiql.css" rel="stylesheet">
   {% endblock style %}
+    <script src="https://unpkg.com/whatwg-fetch@{{ versions.fetch }}/fetch.js"></script>
```
ooflorent: Could you fix the indent here please?

ooflorent:
Many thanks for this! Looks good to me, we can merge this as soon as the Twig indent is fixed.

pascaldevink:
Thanks for the quick response 😃 

@ooflorent I've rebased the fix, stupid editor decided to indent on it's own!

PR #108, opened 2017-01-27
Optimize react promise

mcg-web:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented? |   no
| Fixed tickets | none
| License       | MIT

Fix memory leak in react adapter wait
Also remove dead code and add react promise composer suggest.

### DIFF
--- a/Executor/Promise/Adapter/ReactPromiseAdapter.php
+++ b/Executor/Promise/Adapter/ReactPromiseAdapter.php
@@ -41,16 +41,17 @@ class ReactPromiseAdapter extends BaseReactPromiseAdapter implements PromiseAdap
     /**
      * Synchronously wait when promise completes.
      *
-     * @param Promise $promise
+     * @param Promise  $promise
+     * @param callable $onProgress
      *
      * @return ExecutionResult
      *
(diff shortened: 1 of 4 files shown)


Executor/Promise/Adapter/ReactPromiseAdapter.php
```diff
@@ -58,17 +58,20 @@ public function wait(Promise $promise)
...
+            $resolvedValue = $values;
+            $wait = false;
+        }, function ($reason) use (&$exception, &$wait) {
+            $exception = $reason;
+            $wait = false;
+        });
+
+        // wait until promise resolution
         while ($wait) {
-            $reactPromise->then(function ($values) use (&$resolvedValue, &$wait) {
```
ooflorent: Add a comment inside this block to explain why

ooflorent: Maybe adding `usleep(1)` would make it less CPU intensive without sacrificing the performance

Issue #12, opened 2016-03-23
GraphQL Type System Definition to a config yaml


mcg-web:
Add a feature that convert  GraphQL [Type System Definition](https://facebook.github.io/graphql/#sec-Type-System-Definition):

```
interface Character {
  id: String!
  name: String
  friends: [Character]
  appearsIn: [Episode]
}
```

to config yaml file:

``` yaml
Character:
    type: interface
    config:
        fields:
            id:
                type: "String!"
            name:
                type: "String"
            friends:
                type: "[Character]"
            appearsIn:
                type: "[Episode]"
```

akomm:
Do you want it as an alternative to yml config, or separate - lets say command - to generate the yml out of schema definition?

mcg-web:
hi @akomm,
the need has change since... Now that **webonyx/graphql-php** supports Type System Definition to schema natively, graphql definition we be an easy way to write the schema. This solution will be more reusable than yaml, xml or php definition...

enumag:
I think I might have found a work-around...

First define a SchemaFactory:

```
<?php declare(strict_types = 1);

namespace App\GraphQL;

use GraphQL\Type\Schema;
use GraphQL\Utils\BuildSchema;

class SchemaFactory
{
	public static function create(string $rootDir): Schema
	{
		// TODO: Add caching according to http://webonyx.github.io/graphql-php/type-system/type-language/#performance-considerations
		$contents = file_get_contents($rootDir . '/config/graphql/types/schema.graphql');

		return BuildSchema::build($contents);
	}
}
```

Then register the schema in services.yml:

```
    overblog_graphql.schema_default:
        factory: 'App\GraphQL\SchemaFactory:create'
        arguments:
            - %kernel.project_dir%
```

And add the schema.graphql file:

```
schema {
    query: Query
}

type Query {
    greetings(input: HelloInput!): String!
}

input HelloInput {
    firstName: String!
    lastName: String
}
```

What I don't understand however is how to define the resolver needed for the `greetings` query to work. Currently it throws this error:

```
Cannot return null for non-nullable field Query.greetings.
```

I know why, I didn't define anywhere how the greetings query should respond after all. But how do I fix it?

mcg-web:
[...]

PR #113, opened 2017-02-11
Allow GraphQL schema shorthand configuration

jane-olszewska:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | #12
| License       | MIT

This PR adds the ability to define your schema in GraphQL schema shorthand.

It requires library version updates (patch version level) for better comment handling. Comments are the mechanism by which you can define elements that aren't reflected in schema shorthand: descriptions, resolvers and custom enum value values.

WIP - "men at work" warning - assume this branch will be rebased.

TODO, reminder to self:

- [ ] clean up `OverblogGraphQLTypesExtension::parseSchemaToOverblogGraphQLTypesConfigArray()`
- [ ] add unit tests (loading of the new configuration type etc)
- [ ] add ability to load multiple types of configuration files from the same directory
- [ ] add documentation with examples (esp. resolvers and custom enum value values)
- [ ] add credits on collaborated code
- [ ] improve commit messages where needed

### DIFF
--- a/Definition/Builder/SchemaBuilder.php
+++ b/Definition/Builder/SchemaBuilder.php
@@ -13,6 +13,7 @@ namespace Overblog\GraphQLBundle\Definition\Builder;
 
 use GraphQL\Schema;
 use GraphQL\Type\Definition\Config;
+use GraphQL\Type\LazyResolution;
 use Overblog\GraphQLBundle\Resolver\ResolverInterface;
 
 class SchemaBuilder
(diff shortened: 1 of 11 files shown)


mcg-web:
Hi @jane-olszewska, 
Thank you for the contribution, I think that we should keep the schema totally compatible with the specs so doing leaves the possibility to use it with other servers if needed. If builders are not supported by this implementation, that's not a big deal. For resolvers, we can maybe introduce `resolverMap` like [apollographql/graphql-tools](https://github.com/apollographql/graphql-tools)? For the parser it may be time to introduce dedicated parser for each type...
That a huge work, tell me what you think of all this.

jane-olszewska:
Hi @mcg-web :)

> I think that we should keep the schema totally compatible with the specs so doing leaves the possibility to use it with other servers if needed.

At the moment, it is compatible in that it won't break validation when moved over to different servers (and it's very easy to strip out by matching to a regexp). (In comparison to other extensions, a bit more compatible than for example + based syntax for decorators as introduced in the apollo extension.)

> If builders are not supported by this implementation, that's not a big deal. For resolvers, we can maybe introduce resolverMap like apollographql/graphql-tools?

The resolvers are already declared as services and tagged appropriately so maybe it would be better to add custom attributes to service tags (a combination of type/field)? But the most convenient option would be to be able to jump from the type definition straight to the resolver method.

> For the parser it may be time to introduce dedicated parser for each type...

WDYM?

> That a huge work, tell me what you think of all this.

Yeah, so this PR is an extract of code I wrote for another project as you can tell by the 'todo' comments in the code (that will be cleaned up). Unfortunately, I don't currently have a lot of time to spend on expanding the library, but I thought people might make use of it as it is, even if the functionality is a bit limited.

mcg-web:
I think this new feature of webonyx/graphql-php could help a lot on this webonyx/graphql-php#91...

PR #123, opened 2017-03-17
Regression: Fix graphiql route when running in the prod env

adri:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | -
| License       | MIT

When running in `prod` the `_wdt` route is not available.
Leads to this PHP error:

```
Uncaught PHP Exception Twig_Error_Runtime: "An exception has been thrown during the rendering of a template ("Unable to generate a URL for the named route "_wdt" as such route does not exist.").
```

 Regression introduced in [this commit](https://github.com/overblog/GraphQLBundle/commit/e5ae4ed0cc13df67af7f04544dcbe88d64e4805b). 

PS: I love this bundle!

### DIFF
--- a/Resources/views/GraphiQL/index.html.twig
+++ b/Resources/views/GraphiQL/index.html.twig
@@ -46,7 +46,7 @@
         var xdebugToken = res.headers.get('X-Debug-Token')
         if (typeof Sfjs !== "undefined" && xdebugToken) {
           var toolbarElement = document.querySelector('.sf-toolbar')
-          var debugUrlPattern = "{{ url('_wdt', {'token': '__TOKEN__'}) }}"
+          var debugUrlPattern = "{{ app.debug ? url('_wdt', {'token': '__TOKEN__'}) : '/_wdt/__TOKEN__' }}"
           Sfjs.load(toolbarElement.id, debugUrlPattern.replace('__TOKEN__', xdebugToken ))
         }
         return res.text()
--- a/Resources/views/GraphiQL/index.html.twig
+++ b/Resources/views/GraphiQL/index.html.twig
@@ -43,12 +43,14 @@
         body: JSON.stringify(params),
         credentials: 'include',
       }).then((res) => {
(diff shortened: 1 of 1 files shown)


mcg-web:
Thanks @adri for reporting this and for the **PS**!

Resources/views/GraphiQL/index.html.twig
```diff
@@ -46,7 +46,7 @@
         var xdebugToken = res.headers.get('X-Debug-Token')
         if (typeof Sfjs !== "undefined" && xdebugToken) {
           var toolbarElement = document.querySelector('.sf-toolbar')
-          var debugUrlPattern = "{{ url('_wdt', {'token': '__TOKEN__'}) }}"
+          var debugUrlPattern = "{{ app.debug ? url('_wdt', {'token': '__TOKEN__'}) : '/_wdt/__TOKEN__' }}"
```
mcg-web: maybe the best way is to remove the `if` condition if not in debug mode? so doing we don't keep useless condition.
what you think of that?

adri: Yeah we can wrap the `if` (javascript) in a `if app.debug` (twig) :) 

The reason why I didn't do it was because `app.debug` doesn't necessarily mean that there is no web debug toolbar. I was trying to find a good way to test for that but didn't see a good way. Best would be IMO to somehow test for that. Do you have an idea?

mcg-web: Checking that `_wdt` route existing could be the right way to that. [This could maybe help to do that.](http://stackoverflow.com/questions/17241235/test-if-route-exists-in-twig-template-symfony-2)

adri:
@mcg-web After looking into it for a bit I found that it's just easier to test for `app.debug` :)

mcg-web:
Thank you @adri!

adri:
Thanks! :)

PR #125, opened 2017-03-28
Add a 'public' property to control field visibility

alafon:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? |no
| Tests pass?   | yes
| Documented?   | not yet
| Fixed tickets | n/a
| License       | MIT

Make it possible to completely remove a field (from any queries, including inspection queries) based on the Expression Language, by not exposing it.

When defining fields, `public` works the same way as `access`, except that defining a default `public` behavior is not supported because it does not make sens.


Resources/skeleton/TypeSystem.php.skeleton
```diff
@@ -27,4 +27,19 @@
 <spaces><spaces>$class = substr(strrchr(__CLASS__, "\\"), 1);
 <spaces><spaces>return substr($class, 0, -4);
 <spaces>}
+
+<spaces>private static function applyExposeFilters($fields)
```
mcg-web: Why this is static function?

Generator/TypeGenerator.php
```diff
@@ -72,6 +82,34 @@ protected function resolveTypeCode($alias)
         return  sprintf('$container->get(\'%s\')->resolve(%s)', 'overblog_graphql.type_resolver', var_export($alias, true));
     }
 
+    protected function generateExpose(array $value)
+    {
+        $exposeIsSet = $this->arrayKeyExistsAndIsNotNull($value, 'expose');
+        $fieldOptions = $value;
+        if (!$this->arrayKeyExistsAndIsNotNull($fieldOptions, 'expose')) {
+            $fieldOptions['expose'] = function () {
```
alafon: `return 'null'` instead of the callback function

Generator/TypeGenerator.php
```diff
@@ -62,6 +62,16 @@ protected function generateClassDocBlock(array $value)
 EOF;
     }
 
+    protected function generateOutputFields(array $config)
+    {
+        $outputFieldsCode = sprintf(
+            '$this->applyExposeFilters(%s)',
```
alafon: Use `self::` instead of `$this`

alafon:
- comments from @mcg-web taken into account
- commits rebased
- doc added

With ❤️

ooflorent:
I don't like the naming. Could you rename the field `public` to match Symfony's DI config?

ooflorent:
Another issue is schema generation. If you want to use Relay, use MUST generate the schema upfront. Private fields will then be an issue...

alafon:
@ooflorent do you think that it's a blocker for the PR to be merged? If you think of a specific use case where it would lead to unwanted situation please let me know.

mcg-web:
No this is not a blocker to me... can you squash your commits please? then i'll merge this.

alafon:
Done, thanks 👍

PR #187, opened 2017-09-18
Compatible with PHP 7.2+ and PHPUnit 6+

renatomefi:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| License       | MIT

The current nightly building is failing because on PHP 7.2+ the function each is deprecated, this one is being used on phpunit 4 and 5.
By upgrading phpunit to 6 the build will normally run on php 7.2+
The change which is regards to the TestCase class still works on phpunit 4, thus making sure no BC break is introduced

screenshot from 2017-09-18 14-19-12

### DIFF
--- a/Tests/Definition/ArgumentTest.php
+++ b/Tests/Definition/ArgumentTest.php
@@ -12,8 +12,9 @@
 namespace Overblog\GraphQLBundle\Tests\Definition;
 
 use Overblog\GraphQLBundle\Definition\Argument;
+use PHPUnit\Framework\TestCase;
 
-class ArgumentTest extends \PHPUnit_Framework_TestCase
+class ArgumentTest extends TestCase
 {
     /**
      * @var array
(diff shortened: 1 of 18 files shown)


Tests/Definition/ArgumentTest.php
```diff
@@ -12,11 +12,12 @@
...
 
 use Overblog\GraphQLBundle\Definition\Argument;
+use PHPUnit\Framework\TestCase;
 
-class ArgumentTest extends \PHPUnit_Framework_TestCase
+class ArgumentTest extends TestCase
 {
     /**
-     * @var array
+     * @var array;
```
mcg-web: extra ";"

renatomefi:
@mcg-web thanks! I've rebased the branch since it was outdated and updated it with your comment

PR #190, opened 2017-09-19
Improving Travis build

renatomefi:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| License       | MIT

- Making travis file more readable within `script` test
- Using composer optimized autoloader for faster builds

Suggestions:
- Use `perfer-source` on composer, since we're saving composer cache by using the git repo it'll checkout the code faster, should I do it?
- Wouldn't be better to make the coverage under php 7+ due to performance?
- What about caching .php_cs.cache? It will always check for the updated files anyway

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ matrix:
         - php: 5.5
           env: SYMFONY_VERSION=2.8.*
         - php: 5.6
-          env: SYMFONY_VERSION=3.1.*
+          env: SYMFONY_VERSION=3.1.* PHPUNIT_FLAGS="-d xdebug.max_nesting_level=1000 --coverage-clover=build/logs/clover.xml"
         - php: 7.0
           env: SYMFONY_VERSION=3.2.*
         - php: 7.1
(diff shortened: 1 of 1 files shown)


mcg-web:
Hi @renatomefi ,
> User perfer-source on composer, since we're saving composer cache by using the git repo it'll checkout the code faster, should I do it?
Yes we can do that
> Wouldn't be better to make the coverage under php 7+ due to performance
using php 5.6 is just legacy so ok for php 7+
> What about caching .php_cs.cache? It will always check for the updated files anyway
Nice catche :+1: 

Thanks for contributing to the project!

renatomefi:
@mcg-web it's my pleasure, I use the bundle a lot :)

I've done the changes and the total build time is already faster, when cache is on the master branch it'll be a bit faster! :)

Do you know why hhvm is so slow?

Something else:
I've done a test before with `prefer-lowest` to make sure the minimal requirements are actually safe and sadly many things are breaking, I might work on that another time!

mcg-web:
@renatomefi i don't understand why HHVM is slow. It became slow after using trusty dist...

Issue #189, opened 2017-09-19
Replace Symfony ClassLoader by composer ClassLoader


mcg-web:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | no
| Version/Branch   | 0.9.x

Symfony ClassLoader is deprecated and should be replace by composer autoloader. In Symfony 4 this bundle generated schema types will be persist, so maybe just using the `autoload` composer.json entry will be sufficient and more optimized.

stof:
If you need your own autoloader, you could indeed instantiate a composer ClassLoader (people using a Symfony bundle without having composer are probably too edge case to require supporting this).
Just make sure you don't alter the configuration of the instance available as `$loader = require 'vendor/autoload.php'`. This would break several optimizations done in Composer for the generated class loader. Use a second instance instead.

Relying on the generated Composer autoloader configured through `composer.json` depends on when the classes are generated (if it is too late, it breaks the usage of classmap-authoritative autoloaders). And it may require the user to alter their composer.json.

stof:
If types are generated as classes meant to be committed in the repo, using the composer autoloader is totally fine.
But make sure that the generated files are located in a way allowing to use PSR-4 autoloading (otherwise, dev environments will be a pain if they have to redump the composer classmap everytime they add a new type)

mcg-web:
First, thanks @stof for your help.

Yes the generated types PSR-4 ready combine with a classMap File that is keep up to date.  
Using my own Composer ClassLoader look to be the best way to go, as altering composer.json could be more painful for users... 

Can this example do it?

```php
$loader = new \Composer\Autoload\ClassLoader();
$classMap = 'my/path-to/__classes.map'
$loader->addClassMap($classMap);
// a little optimization
$loader->setClassMapAuthoritative(false);

$loader->register();
```

Thank you again

stof:
@mcg-web yes (except that `$loader->setClassMapAuthoritative(false);` is useless as `false` is the default value).

However, if the generated types are meant to be committed in the project (rather than being generated on cache warmup for instance), you may not need this at all (just instructing users to add the autoloading rule for their types in composer.json)

mcg-web:
Ok, need some more changes and documentation to make types committed, we will take the change of @renatomefi in first place. Thank you again @stof and also @renatomefi...

renatomefi:
[...]

PR #198, opened 2017-09-21
Allow using project composer Classloader and some more optimizations

mcg-web:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | #189 
| License       | MIT


Generator/TypeGenerator.php
```diff
@@ -216,8 +216,14 @@ public function loadClasses($forceReload = false)
...
         if (!self::$classMapLoaded || $forceReload) {
             $classes = require $this->getClassesMap();
-
-            $mapClassLoader = new ClassLoader();
+            /** @var ClassLoader $mapClassLoader */
+            static $mapClassLoader = null;
+            if (null === $mapClassLoader) {
+                $mapClassLoader = new ClassLoader();
+                $mapClassLoader->setClassMapAuthoritative(true);
+            } else {
```
renatomefi: Is it possible to enter this else statement? `$mapClassLoader` seems to be always null to me or am I missing something?

.travis.yml
```diff
@@ -17,7 +17,7 @@ matrix:
         - php: 7.0
           env: SYMFONY_VERSION=3.2.*
         - php: 7.1
-          env: SYMFONY_VERSION=3.3.* SCRUTINIZER=true
+          env: SYMFONY_VERSION=3.3.* TEST_COVERAGE=true
```
renatomefi: cool :+1:

DependencyInjection/Compiler/ConfigTypesPass.php
```diff
@@ -21,9 +21,18 @@ class ConfigTypesPass implements CompilerPassInterface
     public function process(ContainerBuilder $container)
     {
         $config = $container->getParameter('overblog_graphql_types.config');
-        $generatedClasses = $container->get('overblog_graphql.cache_compiler')->compile($this->processConfig($config));
+        $generatedClasses = $container->get('overblog_graphql.cache_compiler')->compile(
```
renatomefi: I think it's a good idea to extract `$container->get('overblog_graphql.cache_compiler')` and check its type just for safety and the IDE will also type hint it nicely

Generator/TypeGenerator.php
mcg-web: No it is null on init because $mapClassLoader is declare as static so after it init it always has the same instance of ClassLoader.

PR #213, opened 2017-10-19
Allow Symfony 4.x as valid Symfony version

pascaldevink:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| License       | MIT

Since Symfony released the first beta for version 4, which is not all that different from version 3.4, this composer change will allow it to work with that version (if the `minimum-stability` is set to `beta` at least).


.travis.yml
```diff
@@ -42,7 +44,7 @@ install: composer update --prefer-source --no-interaction --optimize-autoloader
 
 script:
     - bin/phpunit --debug $( if [ $TEST_COVERAGE = true ]; then echo "-d xdebug.max_nesting_level=1000 --coverage-clover=build/logs/clover.xml"; fi; )
-    - if [ ${TRAVIS_PHP_VERSION} == "7.0" ]; then bin/php-cs-fixer fix --diff --dry-run -v; fi;
+    - if [ ${TRAVIS_PHP_VERSION} == "7.0" ]; then composer require --dev 'friendsofphp/php-cs-fixer:^2.0' && bin/php-cs-fixer fix --diff --dry-run -v; fi;
```
renatomefi: Good idea! Let's make it run on 7.1 instead for speed?

.travis.yml
```diff
@@ -19,20 +19,22 @@ matrix:
...
           env: SYMFONY_VERSION=3.3.* TEST_COVERAGE=true
         - php: 7.1
-          env: GRAPHQLPHP_VERSION=0.10.0
+          env: SYMFONY_VERSION=3.4.* GRAPHQLPHP_VERSION=0.10.0 DEPENDENCIES=dev
         - php: hhvm
+          dist: trusty
+          group: edge
         - php: nightly
-          env: COMPOSER_UPDATE_FLAGS=--ignore-platform-reqs
+          env: SYMFONY_VERSION=4.0.* DEPENDENCIES=dev COMPOSER_UPDATE_FLAGS=--ignore-platform-reqs
```
renatomefi: Let's make a build 7.1 with symfony 4 and leave the nightly more accordingly to our composer.json?

Tests/Functional/App/config/exception/config.yml
```diff
@@ -7,7 +7,7 @@ overblog_graphql:
         class_namespace: "Overblog\\GraphQLBundle\\Exception\\__DEFINITIONS__"
         exceptions:
             errors:
-              - "InvalidArgumentException"
+              - 'InvalidArgumentException'
```
renatomefi: Is this necessary? Feels like we are now mixing single and double quotes, maybe we can choose one

mcg-web: nope no necessary, forget to rollback this

.travis.yml
mcg-web: remove it from requirements because it seems to not be compatible with Symfony 4

.travis.yml
mcg-web: yes, I will make the change

CONTRIBUTING.md
```diff
@@ -16,5 +16,6 @@ Giving some love to PHP CS
 ---------------------------
 
 ```bash
-bin/php-cs-fixer fix ./
+composer require --dev 'friendsofphp/php-cs-fixer:^2.0'
```
renatomefi: @mcg-web how about creating a composer script to install and run the php cs?
```
composer cs-fix
```

Could even do a `composer test` as well for the phpunit call

mcg-web: I'm not against, we will dedicate a PR to this

mcg-web:
@pascaldevink it seems that this is asking some more work to be ready to merge, I'm working on this thank for initializing the work :+1:

renatomefi:
Yeah, I've tried it myself a couple of weeks ago when making the SF Flex thingy, there's indeed some work necessary!

pascaldevink:
I would definitely argue that this requires a new version, this seems to be more than a patch, if only because php 5.5 is no longer supported.

It seems all the tests that fail are for hhvm and php 5.6 with symfony 3.1, is there anything I can do to help fix these?

mcg-web:
The issue with this update is the change of container cache compilation as now each service are become a php class against methods in older versions... This new feature hardly touching the way the tests works right now. Some reverse engineering is necessary to understand this.

PR #219, opened 2017-10-24
cast to array the result of parsing yml and xml.

hgraca:
Previously, if the contents of a type file were all commented out, it would return null which goes against the method signature and would break the call to `$container->prependExtensionConfig()`, in `OverblogGraphQLTypesExtension::prependExtensionConfigFromFiles()`.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | 
| License       | MIT

### DIFF
--- a/Config/Parser/XmlParser.php
+++ b/Config/Parser/XmlParser.php
@@ -43,6 +43,6 @@ class XmlParser implements ParserInterface
             throw new InvalidArgumentException(sprintf('Unable to parse file "%s".', $file), $e->getCode(), $e);
         }
 
-        return $typesConfig;
+        return (array) $typesConfig;
     }
 }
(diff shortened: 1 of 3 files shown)


mcg-web:
thank you @hgraca can you submit this fix on 0.9 branch please?

renatomefi:
Would be good to provide a test for this?

hgraca:
@mcg-web done

@renatomefi yes, but we will need to do some more changes then, because the yaml parser instantiates the real parser and the xml parser makes a static call, so we can not mock those things. Bottom line, I will need more time for this and I don't know when I will have it. :( 

Let me know if you prefer I close the PR and submit another one when I have it all done.

mcg-web:
@hgraca thank you! I'll add tests covering this before merging it, but not a big deal.

hgraca:
:tada:

PR #222, opened 2017-10-25
Add the argument specification in the query expl

hgraca:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | 
| License       | MIT

Add the argument specification in the query example


renatomefi:
Let's not forget this is optional, sometimes you might just inject the whole args and deal with it in the resolver, maybe both examples would be good

hgraca:
:+1:

PR #224, opened 2017-10-26
Finding and testing lowest possible versions

renatomefi:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| License       | MIT

- Trying to find out the lowest possible versions the bundle is compatible with
- Slight improves on travis build

----- OLD
- Sadly we don't support expression language 3.1
- Check it's constructor: https://github.com/symfony/symfony/blob/v3.1.10/src/Symfony/Component/ExpressionLanguage/ExpressionLanguage.php#L38
- Now on version 3.2: https://github.com/symfony/symfony/blob/v3.2.0/src/Symfony/Component/ExpressionLanguage/ExpressionLanguage.php#L40

And that's how we use it, as on 3.2: https://github.com/overblog/GraphQLBundle/blob/v0.9.2/ExpressionLanguage/ExpressionLanguage.php#L26

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -47,7 +47,7 @@
         "react/promise": "To use ReactPHP promise adapter"
     },
     "require-dev": {
-        "phpunit/phpunit": "^4.1 || ^5.5 || ^6.0",
+        "phpunit/phpunit": "^5.5 || ^6.0",
         "react/promise": "^2.5",
         "sensio/framework-extra-bundle": "^3.0",
         "symfony/asset": "^3.1 || ^4.0",
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,9 +15,13 @@ matrix:
     fast_finish: true
     include:
         - php: 5.6
-          env: SYMFONY_VERSION=3.1.*
+          env: COMPOSER_UPDATE_FLAGS=--prefer-lowest SYMFONY_DEPRECATIONS_HELPER=disabled
+        - php: 5.6
+          env: SYMFONY_VERSION=3.1.* SYMFONY_DEPRECATIONS_HELPER=disabled
         - php: 7.0
           env: SYMFONY_VERSION=3.2.*
+        - php: 7.1
+          env: PHP_CS_FIXER=true
         - php: 7.1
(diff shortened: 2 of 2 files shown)


renatomefi:
This is going to be a bit more trick since some tests are trying to force 3.1

mcg-web:
I don't understand why expression language 3.1 is not compatible?

renatomefi:
Because the way we are constructing it on https://github.com/overblog/GraphQLBundle/blob/v0.9.2/ExpressionLanguage/ExpressionLanguage.php#L26

This constructor is only available on 3.2, not on 3.1

mcg-web:
both  constructor are compatible (2 params) we using only the second parameter right now to register providers. If it was not compatible tests would break? I don't get it sorry

renatomefi:
They don't break because currently we are not testing with "--prefer-lowest", and there's a BC trick happening, so we are able to use expression language 3.2 with symfony 3.1.

I found it out while testing --prefer-lowest on the GraphiQL.

If we start doing a prefer-lowest here we are gonna find some nasty symfony breaks and we might have to fix each version

mcg-web:
We should maybe add the flag if this is the case?

renatomefi:
For some reason here it didn't fail on the 3.1, while if I make a functional test on the new GraphiQL bundle it fails on 3.1 and passes on 3.2

composer.json
```diff
@@ -47,7 +47,7 @@
         "react/promise": "To use ReactPHP promise adapter"
     },
     "require-dev": {
-        "phpunit/phpunit": "^4.1 || ^5.5 || ^6.0",
+        "phpunit/phpunit": "^5.5 || ^6.0",
```
renatomefi: Since we use `TestCase` and not `PHPUnit_Framework_TestCase` we aren't compatible with phpunit 4

renatomefi:
@mcg-web I've added the prefer-lowest build, it all looks fine!
I squashed since the history wasn't adding value

PR #226, opened 2017-10-27
Rename .php_cs to .php_cs.dist and ignore .php_cs

nstapelbroek:
Hi ya :raising_hand_man: I saw a small improvement on the php-cs-fixer implementation in the repository.

### What has been done
- The upgrade guide states that it's now recommended to rename your .php_cs to .php_cs.dist so you can allow contributors to keep their own conventions. This PR does that.

### How to test
- Validate if the build succeeds and php-cs-fixer is running


renatomefi:
Thank you @nstapelbroek

PR #233, opened 2017-11-03
move code to `src` and tests to `tests`

hgraca:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Documented?   | no
| Fixed tickets | 
| License       | MIT

Move code to `src` and tests to `tests`, just to follow to usual standard.

There is a test failing but it was failing already in the base branch.


renatomefi:
@hgraca which kind of standard are we talking about?
Many bundles are located without src directory. The tests's one is common tho

hgraca:
My standard 😎😂 (sorry, can't resist the thug joke)
Truth is i never worked on a project where the source code was directly in the project root... so i figured that would be the standard. Nevertheless, consider it a standard or not, i feel its better organization of the code, i can't see any negatives and see the project as something more clean, organized. But that's just me, if the maintainers don't feel that way, im perfectly fine with it :)

mcg-web:
I'm not against this re-organisation (some month ago [this already been submit](https://github.com/mcg-web/GraphQLBundle/tree/monolithic-repo)) but this will add complexity to maintain many branches (0.8, 0.9, master) at the same time.  I think we should wait for the right moment to introduce this new organisation (next major release or when 0.9  will be in security maintenance)

hgraca:
ok, lets close the PR then :)

PR #240, opened 2017-11-10
Overall code and tests improvements

renatomefi:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| License       | MIT

- Just some overall small code improvements of things which I'm finding while digging in the code

### DIFF
--- a/Tests/Error/ErrorHandlerTest.php
+++ b/Tests/Error/ErrorHandlerTest.php
@@ -73,12 +73,11 @@ class ErrorHandlerTest extends TestCase
         $this->assertEquals($expected, $executionResult->toArray());
     }
 
-    /**
-     * @expectedException \Exception
-     * @expectedExceptionMessage My Exception message
-     */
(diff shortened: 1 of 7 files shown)


Tests/Error/ErrorHandlerTest.php
```diff
@@ -220,6 +230,11 @@ public function parentExceptionMappingDataProvider()
                 false,
                 ChildOfInvalidArgumentException::class,
             ],
+            'with $mapExceptionsToParent and no classes' => [
```
renatomefi: @hgraca this covers your `return null` in the parent resolvers: https://github.com/overblog/GraphQLBundle/commit/44bea529ca7d8032f17e0652c83b03ae7b12f031#diff-7a5f7164895da8d47513f7ee5157841dR221

renatomefi: Only now I had time to check it properly, good work :100:

PR #242, opened 2017-11-13
Bugfix: [Relay Pagination] Fixed query limit if no 'first' parameter set

sharif9876:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | 
| License       | MIT

Currently with Relay pagination, if you don't include a `first` parameter in your query, your response will be limited to 1 result.

This is because the logic was, for instance: `$limit + 1`, so even if `$limit` was `null`, we'd still end up limiting to 1.

### DIFF
--- a/Relay/Connection/Paginator.php
+++ b/Relay/Connection/Paginator.php
@@ -68,13 +68,13 @@ class Paginator
 
         // If we don't have a cursor or if it's not valid, then we must not use the slice method
         if (!is_numeric(ConnectionBuilder::cursorToOffset($args['after'])) || !$args['after']) {
-            $entities = call_user_func($this->fetcher, $offset, $limit + 1);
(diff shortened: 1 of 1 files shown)


mcg-web:
Thanks for bringing this up @sharif9876! @alafon can you review this please?

adri:
If you type hint the `limit` param in the callback, this is actually a BC break. 
I had to set `defaultValue` on all `first` arguments now.

mcg-web:
@adri I'm not using this part of the bundle so if you much more accustomed with this can you provide some example please?  If needed we can patch this in current stable version.

adri:
@mcg-web No need to patch I think, just wanted to mention this for future reference. 
Setting a `defaultValue` makes sense anyway :-)

PR #250, opened 2017-11-27
Upgrade 0.11: Versions removal

renatomefi:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| License       | MIT

- Documenting the fact some graphiql versions were removed from the GraphQLBundle

How it looks like: https://github.com/renatomefi/overblog-GraphQLBundle/blob/documentation/upgrade-0.11-js-versions/UPGRADE-0.11.md

PR #251, opened 2017-11-28
Feature/remove relay version option

renatomefi:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| License       | MIT

- I felt like changing how the CLI works, but I think for now its good to maintain the behavior and that's what I've done

### TODO
- [ ] merge #250

### DIFF
--- a/Command/GraphQLDumpSchemaCommand.php
+++ b/Command/GraphQLDumpSchemaCommand.php
@@ -12,25 +12,20 @@ use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
-class GraphQLDumpSchemaCommand extends Command
+final class GraphQLDumpSchemaCommand extends Command
 {
     /** @var ContainerInterface */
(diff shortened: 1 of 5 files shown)


Command/GraphQLDumpSchemaCommand.php
```diff
@@ -12,25 +12,20 @@
 use Symfony\Component\Console\Style\SymfonyStyle;
 use Symfony\Component\DependencyInjection\ContainerInterface;
```
mcg-web: Why this should be final ? I'm not against but think final classes just want to understand why.

renatomefi: It's not a must, but I think it's important to do it when we maintain a library, this way we don't promote users to extend certain things we don't want them to. When you leave it open you create some sort of legacy and can make some users suffer to update.
Now if they want to extend it they'll have to create a new command with their behavior, which I think is for their and our advantage!

mcg-web:
Thank you @renatomefi !

PR #259, opened 2017-12-13
Add description field to custom-scalar type

casskir:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | 
| License       | MIT

When a new custom-scalar type defined we can't add description field to it. This patch add current functionality.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ matrix:
         - php: 7.2
           env: SYMFONY_VERSION=3.3.*
         - php: 7.2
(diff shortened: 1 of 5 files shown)


mcg-web:
Thank you @casskir :+1:! I think the test is not covering this case. This should failed because the generator lib is not up to date. I'll add a test for this.

PR #271, opened 2018-01-09
Update « extends » to « inherits » in type inheritance documentation

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | x
| Fixed tickets | #... 
| License       | MIT

The old key `extends` was still used in an example.
And since it was removed for the key `inherits`, I just updated the example :man_shrugging:

### DIFF
--- a/Resources/doc/definitions/type-inheritance.md
+++ b/Resources/doc/definitions/type-inheritance.md
@@ -19,7 +19,7 @@ ObjectA:
     type: object
     # ObjectB inherited config (fields, args...) from ObjectA
     heirs: [ObjectB]
-    extends: [InterfaceA, ObjectC]
+    inherits: [InterfaceA, ObjectC]
 
 ObjectC:
     type: object


akomm:
Wow :+1: I just noticed that the option extends does not exist and it was "extends" in the config. Then opened  the site again and it was "inherits". I thought I am crazy, but just 2 minutes ago this commit fixed it :D

akomm:
You fixed it like 4 minutes after I found the issue ^^

Kocal:
Ahah @akomm me too, I was insane and then I see that it was a small mistake :rofl:

akomm:
Well I noticed the problem of non-existing option. So I checked, whether I have to correct version. I did. then I read the documentation again and noticed the "inherits" mentioned at the end and assumed correctly, that it is just a naming issue on "hot"-code.

I wanted t o write an issue and make PR for this, but when referencing the issue it all was suddenly "inherits" everywhere. That is what first driven me crazy :D . Then I checked last commit and it was "2 minutes ago". :D

PR #276, opened 2018-01-16
Pre resolver event

vbelorus:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no (some functional tests not passed, but seems its not connected with new functionality)
| Documented?   | yes
| License       | MIT

Added new event graphql.pre_resolver. It can be useful for custom annotation processing right before resolver will be called.


mcg-web:
Hi @vbelorus thanks for contribution. This change can severally impact performance for large schema with many resolver. I also don't understand what is the real use case of this, can you provide an example please?

vbelorus:
Hi @mcg-web 
We have **access** property in graphql schema but it will not create global exception with access denied status. And, more importantly, that **access is checked after after all data were fetched**.
The first idea was to create custom annotations and add this annotations to resolver methods and check security settings before resolver.

But now I'm starting to doubt in my solution :)

I found that **public** property processed before resolver and can be used for access control for queries. Maybe in this case my changes is really don't have real use cases (except of reading custom annotations of resolver methods).

mcg-web:
Yeah, **access** is for ACL on runtime (the schema is not modified whatever the user and his permissions), **public** modifies the config of each type before instancing them, so this mean each user have his own schema. When using **public** your client (relay, Apollo, ...) has to use introspection query before querying to achieve the right schema for user, this make caching (client side) more difficult.

vbelorus:
@mcg-web 
After discussions and practice of using this event I decided to close this pull request. It's really don't have real use cases.

PR #330, opened 2018-05-23
Example with a PHP constant

sylfabre:
Example a ENUM type with a PHP constant


mcg-web:
Thank you @sylfabre :+1:

PR #333, opened 2018-05-29
Remove auto mapping configuration

mcg-web:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | none
| License       | MIT

Auto mapping configuration has been remove in favor of the Symfony 4+ service configuration.
**TODO**: Remove config entry from flex recipe for 0.12


UPGRADE-0.12.md
```diff
@@ -0,0 +1,37 @@
...
+       
+           _instanceof:
+               GraphQL\Type\Definition\Type:
+                   tags: ['overblog_graphql.type']
+               Overblog\GraphQLBundle\Definition\Resolver\ResolverInterface:
+                   tags: ['overblog_graphql.resolver']
+               Overblog\GraphQLBundle\Definition\Resolver\MutationInterface:
+                   tags: ['overblog_graphql.mutation']
+       
+           App\GraphQL\:
```
renatomefi: What's the idea of this one?

ogizanagi:
> That looks amazing, should we put a new recipe on flex which already does the configuration for the types at least?

This is unlikely to work in a recipe as the base Kernel class provided by the fwb recipe does load packages di config first, and then `services.yaml`:

https://github.com/symfony/recipes/blob/39244360f0f7c63437fb816be9189610542fc45b/symfony/framework-bundle/3.3/src/Kernel.php#L47-L48

which means definitions would be overwritten by:

https://github.com/symfony/recipes/blob/39244360f0f7c63437fb816be9189610542fc45b/symfony/framework-bundle/3.3/config/services.yaml#L16-L17

mcg-web:
Thanks for the tips @ogizanagi :+1: , well will just remove config entry in 0.12 recipe and leave programmers follow documentation if needed.

UPGRADE-0.12.md
renatomefi: @mcg-web ping here :D

mcg-web: It is the entry where the DI will actually look for the classes. But I'm not sure I'm really understanding your question.

renatomefi: I'm afraid this sample is implying people have a GraphQL namespace within their source, which we don't actually know. So I'm asking to know if there's something else related to this namespace which I might have missed

mcg-web: I should modify to `resource: src/GraphQL` instead as this was the default dir of auto mapping. What do you think of that, is it more explicit?

renatomefi: Maybe we don't even need this declaration, or is necessary for something?

PR #339, opened 2018-06-01
abstract resolver for access control

sylfabre:
Adds `isGranted` and `denyAccessUnlessGranted` to a base resolver class to check access like with Symfony controllers

### DIFF
--- a/Definition/Resolver/AbstractResolver.php
+++ b/Definition/Resolver/AbstractResolver.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace Overblog\GraphQLBundle\Definition\Resolver;
+
+use Overblog\GraphQLBundle\Error\UserWarning;
+use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;
+
+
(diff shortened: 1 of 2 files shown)


sylfabre:
@renatomefi The main idea was to provide a base class to extend from like the AbstractController class in Symfony but I can refactor to make it a Trait with the Container Aware thing and people can decide to have their own base class using the trait.

mcg-web:
A trait (`AcessResolverTrait` just an example) will be more useful indeed no need to use Container Aware just use [autowiring](https://symfony.com/doc/current/service_container/autowiring.html#autowiring-other-methods-e-g-setters) to set your authorization checker.

renatomefi:
I said container aware as an example, indeed please don't use containers! hehe

> A trait (AcessResolverTrait just an example) will be more useful indeed no need to use Container Aware just use autowiring to set your authorization checker.

Perfect :+1:

mcg-web:
Closing this because of inactivity, feel free to continue discussion if this is still needed

PR #390, opened 2018-09-22
Add CLI command to validate Schema

MichaelDemeyer:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Documented?   | no
| Fixed tickets | 
| License       | MIT

Add CLI command to easily run $schema->assertValid()


src/Command/ValidateCommand.php
```diff
@@ -0,0 +1,53 @@
...
+        parent::__construct();
+        $this->typeResolver = $typeResolver;
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->setName('graphql:validate')
+            ->setDescription('Validate schema')
+        ;
```
mcg-web: The bundle allow multiple schema definition so this should also allow it, [see this command](https://github.com/overblog/GraphQLBundle/blob/master/src/Command/GraphQLDumpSchemaCommand.php#L40-L45) for a concrete example.

src/Command/ValidateCommand.php
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): void
+    {
+        $query = $this->typeResolver->resolve('Query');
```
mcg-web: You should not assume that all Query or Mutation are named so. [See how to get an existing scheme](https://github.com/overblog/GraphQLBundle/blob/master/src/Command/GraphQLDumpSchemaCommand.php#L108).

src/Command/ValidateCommand.php
```diff
@@ -0,0 +1,54 @@
...
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): void
+    {
+        $schemaName = $input->getOption('schema');
+        $schema = $this->getRequestExecutor()->getSchema($schemaName);
+
+        try {
+            @$schema->assertValid();
```
mcg-web: Can you comment why are you using `@` here please?

MichaelDemeyer: I rework this one to use a cleaner and more explicit method :
Warning::suppress(true);

src/Command/ValidateCommand.php
```diff
@@ -0,0 +1,57 @@
...
+final class ValidateCommand extends Command
+{
+    use RequestExecutorLazyLoaderTrait;
+
+    private $typeResolver;
+
+    public function __construct(TypeResolver $typeResolver)
+    {
+        parent::__construct();
+        $this->typeResolver = $typeResolver;
```
mcg-web: It seems you don't used this anymore, I'm a right? Can you remove it and construct if this is the case.

MichaelDemeyer: Right, I removed it.

Issue #379, opened 2018-09-10
Log graphql Errors from results


mauriau:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | no
| Version/Branch   | 0.11

somethings like this
I will make the PR

```
<?php

declare(strict_types=1);

namespace Overblog\GraphQLBundle\EventListener;

use Overblog\GraphQLBundle\Error\ErrorHandler;
use Overblog\GraphQLBundle\Event\ExecutorResultEvent;
use Psr\Log\LoggerInterface;

final class ErrorHandlerListener
{
    /** @var ErrorHandler */
    private $errorHandler;

    /** @var bool */
    private $throwException;

    /** @var bool */
    private $debug;
    
    /** @var LoggerInterface */
    private $logger;

    public function __construct(
        ErrorHandler $errorHandler,
        LoggerInterface $logger,
        bool $throwException = false,
        bool $debug = false
    ) {
        $this->errorHandler = $errorHandler;
        $this->throwException = $throwException;
        $this->debug = $debug;
        $this->logger = $logger;
    }

    public function onPostExecutor(ExecutorResultEvent $executorResultEvent): void
    {
        $result = $executorResultEvent->getResult();
        $this->errorHandler->handleErrors($result, $this->throwException, $this->debug);
        $result = $result->toArray();

        if (isset($result['errors'])) {
            $this->logger->error(__METHOD__.' : '.\json_encode($result['errors']));
        }
    }
}

```

PR #396, opened 2018-10-01
Log errors from graphQl result

mauriau:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Documented?   | no
| Fixed tickets | #379 
| License       | MIT

Just add a logger to catch errors from graphQl results


mcg-web:
Thank you @mauriau for contributing, I'm not really sure that user error should be log backend side as errors. Maybe to make this more friendly we should have a config to choose the log level  (default: debug or notice), what do you think of that?

PR #397, opened 2018-10-02
Add content to body with OPTIONS verb

ptondereau:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

OPTIONS http request is used by browsers to fetch CORS from server. I would like to change the content of body because I think an empty response is a bad behavior.
We use Cloudflare in front of our API and it considers that empty body response is an error.
I'm not sure about what we need to send instead of an empty response and I suggest an empty array.


mcg-web:
Thank you @ptondereau :+1:

Issue #349, opened 2018-06-18
[FEATURES] Generate type from entity


CocoJr:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | no
| Version/Branch   | *

Hi,

Thanks a lot for your bundle, it's amazing !
However, i see a features not included in your bundle, but really appreciate i think: just one command to generate all types from the entities.  
  
This is an example of command, based on SF4 and not configurable, just for the example:  
  
```php
<?php
namespace App\Command;

use Doctrine\ORM\EntityManagerInterface;
use Symfony\Component\Console\Command\Command;
use Symfony\Component\Console\Input\InputArgument;
use Symfony\Component\Console\Input\InputInterface;
use Symfony\Component\Console\Output\OutputInterface;
use Symfony\Component\Yaml\Yaml;

class CreateEntitySchemaCommand extends Command
{
    /** @var EntityManagerInterface */
    protected $entityManager;

    /**
     * DebugCommand constructor.
     *
     * @param DataforseoCommonService $dataforseoService
     * @param EntityManagerInterface  $entityManager
     */
    public function __construct(EntityManagerInterface $entityManager)
    {
        $this->entityManager = $entityManager;
        $this->entityManager->getConnection()->getConfiguration()->setSQLLogger(null);

        parent::__construct();
    }

    /**
     * Configure
     */
    protected function configure()
    {
        $this
            ->setName('app:graphql:generate:schema')
            ->addArgument('entitiesDirectory', InputArgument::OPTIONAL, '', __DIR__.'/../Entity/')
            ->setDescription('Generate schema entity');
    }

    /**
     * Execute cmd
     *
     * @param InputInterface $input
     * @param OutputInterface $output
     *
     * @return int|null|void
     *
     * @throws \Exception
     */
    protected function execute(InputInterface $input, OutputInterface $output)
    {
        $configTypeDir = __DIR__.'/../../config/graphql/types/';
        $directory = $input->getArgument('entitiesDirectory');
        $scanned_directory = array_diff(scandir($directory), array('..', '.'));

        foreach ($scanned_directory as $file) {
            if (strpos($file, '.php')) {
                $fileContent = file_get_contents($directory.$file);
                if (!strpos($fileContent, '@ORM\Entity')) {
                    continue;
                }

                $entityName = substr($file, 0, -4);
                preg_match('#namespace (.+);#', $fileContent, $namespace);
[...]

PR #407, opened 2018-10-16
Fields auto-guessing from Doctrine ORM annotations

Vincz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | #349 
| License       | MIT

Annotation GraphQL type guessed from Doctrine ORM annotations


Vincz:
Hey @mcg-web !
Here is the second step. Field type can be auto-guess from Doctrine annotations.
I add to add an other interface for Parser as I needed to first parse file in order to generate a map between the PHP classes and the GraphqlType. 
I'd also like to be able to set configuration on Parser, so I could eventually give additional mapping for Doctrine Type => Graphql Type. But the parsers usage is static, so I'm not sure how to do it without changing a lot.

mcg-web:
Hey Vincz, do not have time to review your PR, I will try to do that before the weekend.

Vincz:
Hi @mcg-web, it's ok, take your time. I still need to work on this anyway. I'll let you know when I'm done.

Vincz:
Hey @mcg-web :)

So, I managed to do almost everything I wanted:
- Auto-guessing type from doctrine annotations
- Auto-guessing query/mutation args and type from php type hint
- Query & Mutation providers
- Input builder (transform array of data into corresponding type class instance) + validation

The last thing I'd like to check is the error handling when validation error occurs in the Input Builder.
I tried to explain everything in the documentation. 

Waiting for your feedbacks :)

Vincz:
Hey @mcg-web :)
So, I made an update and change the name of the "Input Builder" to "Arguments transformer".
I also refactored the way it works in order to be able to validate all arguments of a query (instead of failing the query at first invalid argument) and send related errors. I also add an error formatter for this, but I'm not sure I did it right.

With the Arguments transformer, the idea is to be able to transform a query/mutation data, into instance of classes (with asserts annotation) and then be able to call the validation service on them.

If the validation failed, the resolver is not called as an exception is throw. The exception used is a `InvalidArgumentsError`. The validation errors should be send back to the client. So, I added the error formatter listerner to add a `state` property in the response, with the validation errors (inspired by https://medium.com/@tarkus/validation-and-user-errors-in-graphql-mutations-39ca79cd00bf).

It works fine, but I'm curious about something and can't find any documentation about it.
When I send multiple mutations, like:
```graphql
{
mut1: mutate(1)
mut2: mutate(2)
mut3: mutate(3)
}
```

If I don't have any error, I get an expected  array like :
```json
{
  "data": {
    "mut1": "result",
    "mut2": "result",
    "mut3": "result"
  }
}
```

But if one or more mutation failed, I only get one Error message (with the error for the first failing mutation). I don't get the errors of the others calls, and I don't get the data either for the mutation without error. I'm using Graphiql to test.
What is the supposed logic when multiple mutations are send and one of them fail ? Is it supposed to execute them in order and stop as soon as one fail ?

mcg-web:
The specs says that mutation will be executed in serial order. if a mutation fails all other mutations a not execute, this is not the case for query.

mcg-web:
I just think of something can we separate doctrine annotations and the bundle annotations in two different parser so we can move this part later in a dedicated component an ease future maintains. @Vincz does this make sense to you?

Vincz:
Hey @mcg-web !
It's weird, it seems to me that even if the first mutation failed, the other ones are still executed. I'll check.

About the separation between the doctrine annotations and the bundle annotations, I'm not sure.
At the moment, the doctrine annotations just help to guess a GraphQL field type based on a Mapping between Doctrine types and Graphql types, that's all. And it can only work in conjonction with the annotations parser as it's the only parser working with PHP classes. 
In the parser, we are litteraly reading the Doctrine annotations. We don't read the doctrine metadata of the class (as they are not available at this moment). So it only works if the user has made his Doctrine configuration with annotations (and not we Yaml or Xml or ...).
And it's not the only way to guess types, we also use PHP type hint for example.
So, GraphQL type guessing based on Doctrine annotations, it's just one of the features that comes with the usage of the Annotations parser. 
It feels a bit odd, cause using the annotations parser over the others parsers add a lot of new features, but it's because annotations are tied to PHP classes while other parsers are not. 
Usually, with Doctrine, or the Validation components, the configuration is related to PHP classes and their properties and their methods. So annotations or Xml or Yaml generate the same metadata at the end. But here, it's not the case, we generate configuration for GraphQL objects. So we don't have a link between PHP classes and the generated configuration as we don't have a 1-1 relationship between a PHP classe and a GraphQL type, except with the annotations parser.
I don't know, maybe all of this "annotations stuff" should be an entirely different bundle, as the concept is more "1 GraphQL Object = 1 PHP class", but it would be sad as you did a fantastic job in this bundle.

Vincz:
Hey @mcg-web :)

It's weird, it seems to me that even if the first mutation failed, the other ones are still executed. I'll check.

About the separation between the doctrine annotations and the bundle annotations, I'm not sure.
At the moment, the doctrine annotations just help to guess a GraphQL field type based on a Mapping between Doctrine types and Graphql types, that's all. And it can only work in conjonction with the annotations parser as it's the only parser working with PHP classes. 
In the parser, we are litteraly reading the Doctrine annotations. We don't read the doctrine metadata of the class (as they are not available at this moment). So it only works if the user has made his Doctrine configuration with annotations (and not we Yaml or Xml or ...).
And it's not the only way to guess types, we also use PHP type hint for example.
So, GraphQL type guessing based on Doctrine annotations, it's just one of the features that comes with the usage of the Annotations parser. 
It feels a bit odd, cause using the annotations parser over the others parsers add a lot of new features, but it's because annotations are tied to PHP classes while other parsers are not. 
Usually, with Doctrine, or the Validation components, the configuration is related to PHP classes and their properties and their methods. So annotations or Xml or Yaml generate the same metadata at the end. But here, it's not the case, we generate configuration for GraphQL objects. So we don't have a link between PHP classes and the generated configuration as we don't have a 1-1 relationship between a PHP classe and a GraphQL type, except with the annotations parser.
I don't know, maybe all of this "annotations stuff" should be an entirely different bundle, as the concept is more "1 GraphQL Object = 1 PHP class", but it would be sad as you did a fantastic job in this bundle.

mcg-web:
Ok @Vincz  so we should maybe split the parser like GraphQL parser to make it a little more readable but this can be made after :+1:

Vincz:
Ok @mcg-web, I'll think about it ! But my biggest issue to have a clean parser was more about the fact it's static and that is handling one file at a time (that's why I needed to add an extra interface "PreParserInterface".
I'm currently using the bundle with the new annotations features on one of my projects to see how it goes. It is supposed to be stable, so if you want to have a look on your side and give me your feedbacks :)

mcg-web:
We can change the static way of using parser. I will give this a try as soon I get some time out @Vincz . Thanks again for this contribution!

Vincz:
Sounds good @mcg-web . On my side, I'll continue to use the annotation stuff and push fixes.
So far, it's really nice to use and allow to really quickly get a GraphQL api running.
I'll tell you when it's ready to be merge.

Vincz:
Hey @mcg-web ! I'll add missing tests for the coverage soon and let you know.

Vincz:
Hey @mcg-web :) Everything is green now !
I'll let you check all of this and feel free to contact me in case you need more details about the implementation. 
Have a great christmas !

mcg-web:
Thanks for this awesome contribution, I'll take some time to review it in the coming days. Merry Xmas to you too!

Vincz:
Thank you ! :) 
I'm using it on my projects and so far it is pretty easy to use.
It would be nice if someone else could try it to detect use cases I didn't think about. 
The next step is using the fields builder (from #438), to try to ease the generation of cruds. 
I don't know when the release of the 0.12 is planned.
Feel free to try it and give me your feedback !

CocoJr:
Of course i want to try :P
I created an issue about the mapping with the doctrine annotation and i'm waiting this features to use for my personnal project, but also in my company :)

Be sure, this week-end, i try it with pleasure :)

src/Config/Parser/PreParserInterface.php
```diff
@@ -0,0 +1,19 @@
...
+interface PreParserInterface extends ParserInterface
+{
+    /**
+     * @param \SplFileInfo     $file
+     * @param ContainerBuilder $container
+     * @param array            $configs
+     *
+     * @return mixed
+     */
+    public static function preParse(\SplFileInfo $file, ContainerBuilder $container, array $configs = []);
```
mcg-web: @Vincz I'm preparing the annotation branch to merge it in master but I don't understand why we need this method?

Vincz: Hi! It's for parsers that need to parse files in two times. With the annotations and the doctrine entities converted to types, we first need to have a map of entity => GQL types to handle associations and validation.

mcg-web: Thank you, I will give it a better look to see if we can cache some more data on pre parse to speedup parse process.

Vincz: Hey @mcg-web ! Finally, I'm not very happy with the annotations and the parsing process. 

The preParse allowed me to have the mapping between PHP classes and the GraphQL types in order to auto-guess graphql type from doctrine associations or PHP type hints. So we generate Graphql types / queries from PHP classes or methods. 

So something like that:
```
/**
 * @GQL\Query
 */
public function getItem(int $id) : Item {
    return $this->find($id);
}
```
is automatically turn into a field config like
```
[
    'getItem' => [
        'type' => 'Item',
        'args' => ['id' => ['type' => 'Int!']],
        'resolve' => '@=service(...).getItem(args["id"])'
    ]
]
```
It is possible, because we know that the PHP class `Item` match the Graphql type `Item`.

With the annotations, I also introduced 'Query and Mutation' providers (see [Providers / Query / Mutation](https://github.com/overblog/GraphQLBundle/blob/annotations/docs/annotations/annotations-reference.md#provider)). It's quite useful as it allows to define fields for the root query or mutation from services (as long as the service are accessible in the container through their class name) and having the args and type auto-guessed is really nice too.
To handle this, it's done in the AnnotationParser. When the root query or the root mutation is detected, their fields are also populated from the providers. It works great, but it only works with annotations. If we define the Root Query or Mutation with an other format with a different parser (yaml or xml), it won't work. 

The last feature is the argument transformer (see [Arguments Transformer](https://github.com/overblog/GraphQLBundle/blob/annotations/docs/annotations/arguments-transformer.md)). It also uses the mapping PHP / GQL to be able to transform  Graphql Input into PHP arguments. It's nice because it populates automatically the method arguments and validate them.

At the moment, all of this features are very specific to the annotation parser.

It sucks. 

Maybe we should move all this feature at a higher level. The parsers (all types) would not only return config types but also class mapping and providers, like that for example:

```
Human:
    type: object
    class: App\Entity\Human
```

And it would also be able to return Query or Mutation providers (the properties not defined would be auto-guessed) like that for example:

```
$providers:
    App\Repository\ItemRepository:
        findAll:
            parentType: Query
        findOne:
            parentType: Query
            type: Item
            args: { id: Int! }
            resolve: "@=service("App\Repository\ItemRepository").findOne(args['id'])"
```

So, the parsing process, would become:
1. Parse to get the mapping between PHP classes & GraphQL types.
2. Parse the fields configuration (and allow merge if a type is define in multiple places ?).
3. Parse the providers and add corresponding fields to the parent types.

This way, it will not be specific to annotation anymore.  

Tell me what you think about this !

Vincz: Ok, so for now, I'll just add the fields builders on the annotations and we will be good to go.  We'll check the rest later.

PR #419, opened 2018-11-15
Fix variable name, fixes #418

### LINKED ISSUE #418: Undefined variable: globalVariables
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Version/Branch   | 0.11.9

When using a custom global variable, the following error appears:

```
Notice: Undefined variable: globalVariables
```

In the generate type, we can see this code:

```
'resolve' => function ($value, $args, $context, ResolveInfo $info) use ($globalVariable) {
    return $globalVariables->get('faker')->name;
},
```

The generated code uses `$globalVariables` instead of `$globalVariable`

arnaud-lb:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #418 
| License       | MIT

This fixes #418

### DIFF
--- a/src/DependencyInjection/Compiler/GlobalVariablesPass.php
+++ b/src/DependencyInjection/Compiler/GlobalVariablesPass.php
@@ -32,7 +32,7 @@ final class GlobalVariablesPass implements CompilerPassInterface
                     $expressionLanguageDefinition->addMethodCall(
                         'addGlobalName',
                         [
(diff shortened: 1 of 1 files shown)


mcg-web:
Thank you @arnaud-lb if this fixed a bug introduced in 0.11 this PR should be base on 0.11 branch to be able to create a 0.11 version with this fix. Can please do the change if this is the case?

PR #426, opened 2018-12-03
Fix deprecation on symfony 4.2

chalasr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | n/a
| License       | MIT

Upgrading to sf 4.2, you get the following deprecation:

> A tree builder without a root node is deprecated since Symfony 4.2 and will not be supported anymore in 5.0.
Show context    Hide trace
{▼
  /Volumes/HD/Sites/geolid/lisa/graph/vendor/symfony/config/Definition/Builder/TreeBuilder.php:30 {▶}
  /Volumes/HD/Sites/geolid/lisa/graph/vendor/overblog/graphql-bundle/src/DependencyInjection/Configuration.php:40 {▶}

Here is the fix.

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -37,8 +37,10 @@ class Configuration implements ConfigurationInterface
 
     public function getConfigTreeBuilder()
     {
-        $treeBuilder = new TreeBuilder();
-        $rootNode = $treeBuilder->root(self::NAME);
+        $treeBuilder = new TreeBuilder(self::NAME);
+
+        // BC layer for symfony/config 4.1 and older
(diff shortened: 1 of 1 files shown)


mcg-web:
Thank you for the fix but since this has an impact on the latest stable version, it should be done on 0.11 branch. You can fix cs by running `composer fix-cs` cli. Can you please make these changes?

chalasr:
@mcg-web PR rebased and retargetted, CS fixed also. Thanks.
The build is failing for some reason. It looks like a composer cache issue, could you please have a look?

chalasr:
Looks like, yes. Does not seem useful to me neither

mcg-web:
Thank you @chalasr! @enumag yes this is a duplication, we can remove it indeed.

Issue #437, opened 2019-01-03
Dynamic types / fields generation


Vincz:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | no
| Version/Branch   | 0.12

I was wondering if we could come up with some dynamically generated queries and/or mutations.
With the new annotation system, it's really easy to map Doctrine Entities with GraphQL types.
Now, I'd need to be able to generate, based on config, queries & mutations for my Doctrine Entities.
For example, given an Entity name `MyEntity`, mapped as GraphQL type `MyEntity`, I'd like to generate a root query like `getMyEntities` (resolved by the repository method `findAll`) and `getMyEntity($id)` (resolved by the repository method `find`).
This bundle generate and dump the type classes for us based on configuration.
I'm not sure what would be the best way to achieve something like this:
1. Should we use a "Fields Builder" (kind of the same way the Field Builder work but to generate multiple fields) and add this "FieldsBuilder" to graphql types. 
2. Add an additional PHP Parser to be able to return type (but we wouldn't be able to add fields to types config generated by an other parser).
3. Add a `@GQL\Crud` annotation to be use with the annotations parser (but it would be very specific to annotation & doctrine entities / repositories)
4. Add a Schema Extension system, that would allow to declare extensions able to modify the types configs generated. 

Maybe the solution number 1 is the more flexible and convenient. We would just have to add a property like that:
```
Query:
    type: object
    config:
        fieldsBuilder:
            - class: CrudsBuilder
              config:
                  entities: [MyEntity1, MyEntity2]
        fields:
            ...
```

and for example, a Fields Builder like that:

```
class CrudsFieldsBuilder {
    public function generateFields(array $config) {
        $fields = [];
        foreach($config['entities'] as $entity) {
             $fields['get' . $entity] = [
                 'type' => $entity,
                 'resolve' =>"@=resolver(...)"
                 ...
             ];
        }
        
        return $fields;
    }
}
```

What do you think about this @mcg-web  ?

Vincz:
So, I gave a shot to the solution 1. It's quite simple and working great. It's just about delegating fields configuration generation to an other class. It's the exact same thing as the "Field Builder" except it generate multiple fields instead of just one. I'll clean the code a make a PR.
[...]

PR #438, opened 2019-01-04
Implements Fields builder

Vincz:
| Q             | A
| ------------- | ---
| New feature?  | yes
| BC breaks?    | no
| Tests pass?   | yes
| Documented?   | yes
| License       | MIT

Implement Fields builder, see #437

### DIFF
--- a/docs/definitions/builders/fields.md
+++ b/docs/definitions/builders/fields.md
@@ -1,0 +1,83 @@
+# Fields builder
+
+Builder is a way to don't repeat fields definition.
+
+Define your custom fields builder
+
+```yaml
+#app/config/config.yml
+overblog_graphql:
+    #...
+    definitions:
+        #...
+        builders:
+            fields:
+                - alias: "Timestamped"
+                  class: "MyBundle\\GraphQL\\Fields\\TimestampFields"
+```
+
+Builder class must implements `Overblog\GraphQLBundle\Definition\Builder\MappingInterface`
+
+```php
+<?php
+
+namespace MyBundle\GraphQL\Field;
+
+use Overblog\GraphQLBundle\Definition\Builder\MappingInterface;
+
+class TimestampFields implements MappingInterface
+{
+    public function toMappingDefinition(array $config)
+    {
+        $propertyCreatedAt = isset($config['propertyCreatedAt']) ? $config['propertyCreatedAt'] : 'createdAt';
+        $propertyUpdatedAt = isset($config['propertyUpdatedAt']) ? $config['propertyUpdatedAt'] : 'updatedAt';
+
+        return [
+            'createdAt' => [
+                'description' => 'The creation date of the object',
+                'type' => 'Int!',
+                'resolve' => '@=value.' . $propertyCreatedAt,
+            ],
+            'updatedAt' => [
+                'description' => 'The update date of the object',
+                'type' => 'Int!',
+                'resolve' => '@=value.'. $propertyUpdatedAt,
+            ],
+        ];
+    }
+}
+```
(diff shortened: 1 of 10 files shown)


docs/definitions/builders/fields.md
```diff
@@ -0,0 +1,83 @@
...
+
+```yaml
+User:
+    type: object
+    config:
+        fields:
+            createdAt:
+                description: The creation date of the object
+                type: Int!
+                resolve: "@=value.dateCreated"
```
mcg-web:
```suggestion
                resolve: "@=value.createdAt"
```

docs/definitions/builders/fields.md
```diff
@@ -0,0 +1,83 @@
...
+
+```yaml
+User:
+    type: object
+    config:
+        builders:
+            - builder: Timestamped
+              builderConfig:
+                  propertyCreated: dateCreated
+        fields:
```
mcg-web: This part seem to not be used for example, can we removed it?

src/Config/Processor/BuilderProcessor.php
```diff
@@ -45,6 +48,14 @@
     public static function process(array $configs): array
     {
         foreach ($configs as &$config) {
+            if (isset($config['config']['builders']) && \is_array($config['config']['builders'])) {
+                if (!isset($config['config']['fields'])) {
```
mcg-web: We could merge if not empty and set to builder result otherwise, this remove a useless array merge.

src/Config/TypeWithOutputFieldsDefinition.php
```diff
@@ -16,9 +16,6 @@
     protected function outputFieldsSelection(string $name = 'fields')
     {
         $node = self::createNode($name);
-        $node
```
mcg-web: this should not be removed since builder is process in a normalizer before this part

docs/definitions/builders/fields.md
Vincz: This line is fine, it shows that we can pass config to the fields Builder, so the config is changing the used field.

docs/definitions/builders/fields.md
Vincz: Removed

src/Config/TypeWithOutputFieldsDefinition.php
Vincz: You are right, my mistake, sorry. I put it back and removed the useless code I added to check the fields requirements

src/Config/Processor/BuilderProcessor.php
Vincz: Fixed

src/Config/TypeWithOutputFieldsDefinition.php
```diff
@@ -15,11 +15,9 @@
      */
     protected function outputFieldsSelection(string $name = 'fields')
     {
-        $node = self::createNode($name);
-        $node
+        $node = self::createNode($name)
```
mcg-web:
```suggestion
        $node = self::createNode($name);
        $node
```

Vincz:
Hey @mcg-web ! Any idea on the failed check? Is my php-cs config outdated?
composer fix-cs does not update files on my side.

mcg-web:
@Vincz I think the phpcs.phar v2 has been updated that the reason why this failed, I will fix that directly on master branch. Thanks again for this contribution!

Vincz:
Ok, thank you!
If you don't mind to also merge it in the annotations branch so I can update them, it would be great !
Thanks again :)

PR #442, opened 2019-01-21
change argument check from isset to array_key_exist

osavchenko:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| License       | MIT

Current approach doesn't allow to pass `null` value as a value and when you call `getRawArguments()` you can see `null`-valued parameter, but `offsetExists()` return `false`.


mcg-web:
Hi @osavchenko thanks for this PR but can you please give a concrete use case? I don't really get it right now.

osavchenko:
```$arg->getRawArguments();
// output
array:1 [
  "foo" => null
]

$arg->offsetGet('foo'); // output - false

array_key_exists('foo', $rawArguments); // output - true
```

This is my case

osavchenko:
Yes, sure. Done in #443

PR #463, opened 2019-03-14
Add security args on paginator

Ciloe:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | 
| License       | MIT

Add security to have first or last set in args

### DIFF
--- a/Relay/Connection/Paginator.php
+++ b/Relay/Connection/Paginator.php
@@ -3,6 +3,7 @@
 namespace Overblog\GraphQLBundle\Relay\Connection;
 
 use Overblog\GraphQLBundle\Definition\Argument;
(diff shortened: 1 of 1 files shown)


akomm:
I would follow this:
- [forward](https://facebook.github.io/relay/graphql/connections.htm#sec-Forward-pagination-arguments)
- [backward](https://facebook.github.io/relay/graphql/connections.htm#sec-Backward-pagination-arguments)

Your solution does go in this direction but not fully. 
As per spec, when I got it right, it is rather this: (last && before) || (first && after).

murtukov:
Closing this PR due to the lack of feedback.

Issue #290, opened 2018-02-07
Type-Builder


akomm:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | no
| Version/Branch   | dev-master

Being able to create type builders, like done here:
https://github.com/overblog/GraphQLBundle/blob/bee5671564fe57c10b7686f55ad6e57afd3505fa/Relay/Connection/ConnectionDefinition.php

would save a lot of configuration boilerplate. Currently it is encapsulated inside own _types extension and it seems to me not accessible from outside as consumer.

Example this:

```yml
Mutation:
  someMutation:
    type: 'SomeMutationPayload!'
    args:
      input: {type: 'SomeMutationInput!'}

SomeMutationInput:
  type: 'input-object'
  config:
    fields:
      name: {type: 'String!'}
      # ...
      
SomeMutationPayload:
  type: 'object'
  config:
    fields:
      success: {type: 'SomeMutationSuccessPayload'}
      failure: {type: 'SomeMutationFailurePayload'}

SomeMutationSuccessPayload:
  type: 'object'
  config:
    fields:
      result: {type: 'SomeResultType'}
      # ...

SomeMutationFailurePayload:
  type: 'object'
  config:
    fields:
      _error: {type: 'String'}
      name: {type: 'String'}
```

With the abilities to define types like in the linked config mapper for the relay connection, would enable me to implement a builder for this:
```yml

Mutation:
  someMutation:
    builder: 'my-mutation-builder'
    builderConfig:
      name: 'SomeMutation'
      inputFields:
        name: {type: 'String!'}
        # ...
      successFields:
        result: {type: 'SomeResultType'}
        # ...
      failureFields:
        _error: {type: 'String'}
        name: {type: 'String'}
        # ...
```

mcg-web:
Sorry but I don't really get it, what do you want to do with type builder exactly?

akomm:
Something like this as definition mapper:
```php
use Overblog\GraphQLBundle\Definition\Builder\MappingInterface;

final class MutationDefinition implements MappingInterface
{
    public function toMappingDefinition(array $config)
    {
        $mutationName = $config['name'];
        $inputFields = $config['inputFields'];
        $successFields = $config['successFields'];
        $failureFields = $config['failureFields'];

        return [
            'field' => [
                'type' => $mutationName . 'Payload!',
                'args' => [
[...]

PR #473, opened 2019-03-29
feature #290

akomm:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | #290
| License       | MIT

Todo:
- [x] Implement: Allow `field` builder emit extra types (BC)
- [x] Unit test for the above
- [x] Implement: Allow `fields` builder emit extra types (BC)
- [x] Unit test for the above
- [x] Adjust things according to decisions below
- [x] Tests for api abuse detecting, when builders accidentally override types

Todo decision making (please give your feedback on this):
- [x] Should extra types emitted by builders be allowed to override existing types or not? 
**Current behavior**: override. 
**My suggetion**: Do not allow, throw exception - otherwise hard to debug issues can happen.
- [x] Should extra types emitted by `fields` builders be allowed to use other builders?
**Current behavior**: No, they are not processed. Those  types are "completed" as provided by the builder
**My suggestion**: Leave it this way. Prevents a lot of configuration complexity and handling of infinite recursions is not needed this way.

### DIFF
--- a/src/Config/Processor/BuilderProcessor.php
+++ b/src/Config/Processor/BuilderProcessor.php
@@ -47,19 +47,42 @@ final class BuilderProcessor implements ProcessorInterface
      */
     public static function process(array $configs): array
     {
+        $addedTypes = [];
+        // map: "type name" => "provided by" for better DX, while debugging accidental type overrides in builders
+        $reservedTypesMap = \array_combine(
+            \array_keys($configs),
+            \array_fill(0, \count($configs), 'configs')
+        );
+
         foreach ($configs as &$config) {
             if (isset($config['config']['builders']) && \is_array($config['config']['builders'])) {
-                $buildersFields = self::processFieldsBuilders($config['config']['builders']);
-                $config['config']['fields'] = isset($config['config']['fields']) ? \array_merge($buildersFields, $config['config']['fields']) : $buildersFields;
+                ['fields' => $buildersFields, 'types' => $buildersTypes] = self::processFieldsBuilders(
+                    $config['config']['builders'],
+                    $reservedTypesMap
+                );
+
+                $config['config']['fields'] = isset($config['config']['fields'])
+                    ? \array_merge($buildersFields, $config['config']['fields'])
+                    : $buildersFields;
+
+                $addedTypes = \array_merge($addedTypes, $buildersTypes);
+
                 unset($config['config']['builders']);
             }
 
             if (isset($config['config']['fields']) && \is_array($config['config']['fields'])) {
-                $config['config']['fields'] = self::processFieldBuilders($config['config']['fields']);
+                ['fields' => $buildersFields, 'types' => $buildersTypes] = self::processFieldBuilders(
+                    $config['config']['fields'],
+                    $reservedTypesMap
+                );
+
+                $config['config']['fields'] = $buildersFields;
+
+                $addedTypes = \array_merge($addedTypes, $buildersTypes);
             }
         }
 
-        return $configs;
+        return \array_merge($configs, $addedTypes);
     }
 
     public static function addBuilderClass($name, $type, $builderClass): void
@@ -100,8 +123,10 @@ final class BuilderProcessor implements ProcessorInterface
         }
     }
 
-    private static function processFieldBuilders(array $fields)
+    private static function processFieldBuilders(array $fields, array &$reservedTypesMap)
     {
+        $newTypes = [];
+
         foreach ($fields as &$field) {
             $fieldBuilderName = null;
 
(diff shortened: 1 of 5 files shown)


src/Config/Processor/BuilderProcessor.php
```diff
@@ -126,30 +138,53 @@ private static function processFieldBuilders(array $fields)
...
             if ($fieldBuilderName) {
-                $buildField = self::getBuilder($fieldBuilderName, self::BUILDER_FIELD_TYPE)->toMappingDefinition($builderConfig);
-                $field = \is_array($field) ? \array_merge($buildField, $field) : $buildField;
+                $mapping = self::getBuilder($fieldBuilderName, self::BUILDER_FIELD_TYPE)->toMappingDefinition($builderConfig);
+
+                $fieldMapping = $mapping['field'] ?? $mapping;
+                $field = \is_array($field) ? \array_merge($fieldMapping, $field) : $fieldMapping;
+
+                if (isset($mapping['field'], $mapping['types'])) {
+                    $types = \array_merge($types, $mapping['types']);
```
akomm: Check added types in '$mapping['types']' not in `$types` already. Throw if so, including type name and builder that emitteed the duplicate.

src/Config/Processor/BuilderProcessor.php
```diff
@@ -47,19 +47,29 @@ final class BuilderProcessor implements ProcessorInterface
...
+                ['fields' => $buildersFields, 'types' => $buildersTypes] = self::processFieldBuilders($config['config']['fields']);
+
+                $config['config']['fields'] = $buildersFields;
+
+                $addedTypes = \array_merge($addedTypes, $buildersTypes);
             }
         }
 
-        return $configs;
+        return \array_merge($configs, $addedTypes);
```
akomm: Prevent $addedTypes from overriding types comming from $config. Throw in case of duplicate. At this point we do not know which builder it was. We could pass in to the processField*Builders methods a list of type names comming from $config, so they can check this. So they prevent builders override each others extra types and also the ones from $config and give clear message about what is the case and which builder it was.

akomm:
@mcg-web I would complete this just want a second thought on the decision things. You agree with the override restriction from generated types as I annotated? :) Did not want request a regular review as it is WIP

mcg-web:
Hi @akomm thanks for this PR, I agree with both of you suggestions, look good to me right now.

akomm:
@mcg-web I ralized it is impossible to construct a new `\Overblog\GraphQLBundle\DependencyInjection\OverblogGraphQLExtension` and load it with different configurations.

Its because of static $config in `\Overblog\GraphQLBundle\DependencyInjection\OverblogGraphQLExtension::treatConfigs`. So whenever you create a new one in a test, you have all the stuff from other tests in it. Also when you want add something different, its ignored and you get the first configuration. Is it intended?

mcg-web:
This static config variable was introduced to process the config once on compile time (for prepend and regular config treatment) but this can be removed safely since this will never be used on runtime.

akomm:
@mcg-web do you want this change as a separate PR or can I make it in this PR?

mcg-web:
Not a must but in a separated PR this can be merge faster.

akomm:
Removing the caching actually breaks tests, depending on whether I run the whole test or just a subsection. Which is a sign that some tests maybe worked due to the caching. It looks like now tests started to override each others type configs in Functional tests, while before they did not due to the caching. 

Also I was a bit confused why the reloading in one already existing test worked: `\Overblog\GraphQLBundle\Tests\DependencyInjection\OverblogGraphQLTypesExtensionTest::testCustomExceptions` this is because of @runInseparateProcess. I assume it was done because of the same issue I have.

Therefore I will not remove the caching and instead place the test in a separate process for now.

akomm:
I will document it in a separate PR. Just want it be done for now when everything passes fine. I have to go home :)

src/Config/Processor/BuilderProcessor.php
```diff
@@ -126,30 +151,92 @@ private static function processFieldBuilders(array $fields)
...
+        $newTypes = [];
+
         foreach ($builders as $builder) {
             $builderName = $builder['builder'];
-            $builderConfig = null;
-            if (isset($builder['builderConfig'])) {
-                $builderConfig = $builder['builderConfig'];
+            $builderConfig = $builder['builderConfig'] ?? null;
+
+            $mapping = self::getFieldBuilderMapping($builderName, self::BUILDER_FIELDS_TYPE, $builderConfig, $reservedTypesMap);
```
mcg-web: Scrutinizer
/ Inspection => It seems like ``$builderConfig`` can also be of type ``null``; however, parameter ``$builderConfig`` of ``Overblog\GraphQLBundle\Config\Processor\BuilderProcessor::getFieldBuilderMapping()`` does only seem to accept ``array``, maybe add an additional type check?

src/Config/Processor/BuilderProcessor.php
```diff
@@ -47,19 +47,42 @@ final class BuilderProcessor implements ProcessorInterface
...
+            \array_fill(0, \count($configs), 'configs')
+        );
+
         foreach ($configs as &$config) {
             if (isset($config['config']['builders']) && \is_array($config['config']['builders'])) {
-                $buildersFields = self::processFieldsBuilders($config['config']['builders']);
-                $config['config']['fields'] = isset($config['config']['fields']) ? \array_merge($buildersFields, $config['config']['fields']) : $buildersFields;
+                ['fields' => $buildersFields, 'types' => $buildersTypes] = self::processFieldsBuilders(
+                    $config['config']['builders'],
+                    $reservedTypesMap
```
mcg-web: Scrutinizer
/ Inspection => It seems like ``$reservedTypesMap`` can also be of type ``false``; however, parameter ``$reservedTypesMap`` of ``Overblog\GraphQLBundle\Config\Processor\BuilderProcessor::processFieldsBuilders()`` does only seem to accept ``array``, maybe add an additional type check?

src/Config/Processor/BuilderProcessor.php
akomm: This is right as you say it. I also seen this. The problem was, before it was by default null'ed too and passed to the MappingInterface API which also only accepted array (not nullable). So I was thinking fixing this should be BC because the code that had the null case would not have worked for anybody anyway. Since it was an error unrelated to this PR I also did not want to introduce new exception, as this might be seen as BC break (before no exception, error). Also defaulting it to empty array is BC break I realized, because before it failed with error, but now it would silently pass empty config. So you want type check and exception?

src/Config/Processor/BuilderProcessor.php
akomm: This comes from the `array_combine ` function above. It returns false when `keys` and `values `args do not have same amount of elements. Its in the `::process` method right in the beginning. 

In this case the call to array_combine does not have any inputs which could make this happen. So the problem is: When I make a check and throw, the throw at this moment can not be tested and covered and would result in coverage drop. I also can not default it to empty array or similar, as it would allow to accidentally override types.

I had the same issue at a different place, where I had thrown a LogicException which could only be triggered by chenging const values in private code and could not be triggered in unit test to be covered & tested.

Is there a way to annotate skip coverage line from counting, when code is actually unreachable? Or do you want unreachable leaves to be added at all?

src/Config/Processor/BuilderProcessor.php
mcg-web: ok I'll fix from 0.11 it after merging this

src/Config/Processor/BuilderProcessor.php
mcg-web: not a big deal, if this is a case not possible when using the bundle no need to fix it.

mcg-web:
Thank you for this PR :+1:

PR #481, opened 2019-04-24
Validate configuration for `MutationFieldDefinition`

paulinevos:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| License       | MIT

The reason for this change is that when you use Symfony Expression language to configure mutation types, there is really no validation for it. If, for instance, one just passes a string without the expression trigger `@=`, the string will just be `json_encoded` and you won't end up with a valid callable, i.e.:
```
"mutation('graphql.some_mutation', [value['foo'], value['bar']])"
"{"mutation":"graphql.some_other_mutation"}"
```
(correct), versus:
```
""mutation('graphql.some_mutation', [value['foo'], value['bar']])""
```
(incorrect).

It would then not not be able to resolve the mutation response and complain that it
`Cannot return null for non-nullable field some_mutation.id`, which doesn't tell the developer anything. It took our team a really long time to figure out we were missing the `=` in the expression trigger.

Therefore (as far as I understand it) we should always expect either an expression string or an array. Please let me know if I've misunderstood and we should also accept regular string callables. In that case, we should probably change where it `json_encode`s any string that is not an expression.

Lastly the `MutationFieldDefinition` class was lacking some coverage so that's been improved a bit.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,6 +29,7 @@
     },
     "require": {
         "php": ">=7.1",
+        "ext-json": "*",
         "psr/log": "^1.0",
         "symfony/config": "^3.4 || ^4.0",
         "symfony/dependency-injection": "^3.4 || ^4.0",
@@ -38,7 +39,7 @@
         "symfony/options-resolver": "^3.4 || ^4.0",
         "symfony/property-access": "^3.4 || ^4.0",
         "webonyx/graphql-php": "^0.13.0"
-},
+    },
     "replace": {
         "overblog/graphql-php-generator": "self.version"
     },
--- a/src/Relay/Mutation/MutationFieldDefinition.php
+++ b/src/Relay/Mutation/MutationFieldDefinition.php
@@ -14,13 +14,11 @@ final class MutationFieldDefinition implements MappingInterface
     public function toMappingDefinition(array $config): array
     {
         $mutateAndGetPayload = $this->cleanMutateAndGetPayload($config);
-        $payloadType = isset($config['payloadType']) && \is_string($config['payloadType']) ? $config['payloadType'] : null;
-        $inputType = isset($config['inputType']) && \is_string($config['inputType']) ? $config['inputType'].'!' : null;
 
         return [
-            'type' => $payloadType,
+            'type' => $this->extractPayloadType($config),
             'args' => [
-                'input' => ['type' => $inputType],
+                'input' => ['type' => $this->extractInputType($config)],
             ],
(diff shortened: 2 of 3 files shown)


composer.json
```diff
@@ -29,6 +29,7 @@
     },
     "require": {
         "php": ">=7.1",
+        "ext-json": "^1.7",
```
mcg-web:
```suggestion
        "ext-json": "*",
```
since we don't really depend on a specific version

src/Relay/Mutation/MutationFieldDefinition.php
```diff
@@ -5,36 +5,67 @@
...
-        if (\is_string($mutateAndGetPayload) && 0 === \strpos($mutateAndGetPayload, '@=')) {
-            $cleanMutateAndGetPayload = \substr($mutateAndGetPayload, 2);
-        } else {
-            $cleanMutateAndGetPayload = \json_encode($mutateAndGetPayload);
+        if (is_string($mutateAndGetPayload) && \strpos($mutateAndGetPayload, '@=') === 0) {
+            return;
+        }
+
+        if ($mutateAndGetPayload === null) {
+            throw new InvalidArgumentException('Mutation "' . self::KEY_MUTATE_GET_PAYLOAD . '" config is required.');
```
mcg-web: use sprintf please

src/Relay/Mutation/MutationFieldDefinition.php
```diff
@@ -5,36 +5,67 @@
...
+        }
+
+        if (is_string($mutateAndGetPayload)) {
+            throw new InvalidArgumentException('Cannot parse "' . self::KEY_MUTATE_GET_PAYLOAD . '" configuration string.');
```
mcg-web: same as above

src/Relay/Mutation/MutationFieldDefinition.php
```diff
@@ -5,36 +5,67 @@
...
+        }
+
+        if (! is_array($mutateAndGetPayload)) {
+            throw new InvalidArgumentException('Invalid format for "' . self::KEY_MUTATE_GET_PAYLOAD . '" configuration.');
```
mcg-web: same as above

paulinevos:
@mcg-web Thanks for the speedy review. I've addressed your comments

mcg-web:
@paulinevos Thank you!

paulinevos:
@mcg-web Thanks for the quick review + merge 😄

PR #498, opened 2019-06-03
Fix body parsing for GET requests

runjak:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Documented?   | no
| Fixed tickets | none
| License       | MIT

Hi!

So I was reading the [serving-over-http](https://graphql.org/learn/serving-over-http/) section on graphql.org which states:

> Your GraphQL HTTP server should handle the HTTP GET and POST methods.

So I went ahead and tried using GET requests against a piece of software that uses the GraphQLBundle. For that our code returned `{"message":"The request content body must not be empty when using JSON content type request.","code":400}`.
A colleague and me searched further and ended up at `src/Request/Parser.php:33` and found that `getParsedBody` throws when the body is empty, but does not consider GET requests.

By looking trough the code my colleague found that we could use `$parsedBody = [];` in this case which we tested and confirmed to work on our machines.

I checked the tests and coding style and found that one test is now broken: `testEndpointWithEmptyJsonBodyQuery`. Since the test uses a GET I tried to simply use a POST instead because that is when it *should* break. Alas this turned out to break even more tests.
Sorry, that I haven't fixed this but I have trouble navigating the test structure.

If someone could help with the tests I'd be glad to see this PR going forward.

I hope this is an OK way to address this problem. If not I can happily file an issue instead.

-----

Addendum: I'm also not sure where and how this should be documented, but the initial Q/A table asked for that.

### DIFF
--- a/src/Request/Parser.php
+++ b/src/Request/Parser.php
@@ -35,6 +35,7 @@ class Parser implements ParserInterface
     private function getParsedBody(Request $request): array
     {
         $body = $request->getContent();
+        $method = $request->getMethod();
         $contentType = \explode(';', (string) $request->headers->get('content-type'), 2)[0];
 
         switch ($contentType) {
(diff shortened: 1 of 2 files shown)


mcg-web:
Thank you @runjak for this PR, I'll modify test before merging this. Tests need some refactor to simplify them I agree.

runjak:
Hi @mcg-web - thanks for the update and for making the test modifications 👌

PR #504, opened 2019-06-19
Make resolution error message clearer

theofidry:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | dunno
| Documented?   | no
| Fixed tickets | N/A
| License       | MIT

I had a few types listed as follow:

```yaml
overblog_graphql:
    definitions:
        schema:
            #...
            types:
                - Foo
                # ...
```

But after some refactoring `Foo` has been changed to `Bar`. As a result the type `Foo` no longer existed and could not be loaded. Dumping the schema was resulting in a failure with the obscure error:

```
GraphQL\Error\InvariantViolation: Each entry of schema types must be instance of GraphQL\Type\Definition\Type but entry at 3 is null
```

And the trace was not much more helpful. With the following patch, the error is changed to:

```
LogicException: Could not load the type "Foo"
```

### DIFF
--- a/src/Definition/Builder/SchemaBuilder.php
+++ b/src/Definition/Builder/SchemaBuilder.php
@@ -64,7 +64,7 @@ class SchemaBuilder
             'types' => function () use ($types, $schemaName) {
                 $this->typeResolver->setCurrentSchemaName($schemaName);
 
-                return \array_map([$this->typeResolver, 'getSolution'], $types);
+                return \array_map([$this->typeResolver, 'resolve'], $types);
             },
         ];
     }


mcg-web:
Hi @theofidry, this should already be managed by [these lines](https://github.com/overblog/GraphQLBundle/blob/master/src/Resolver/TypeResolver.php#L70-L74) so we could maybe use `resolve` instead of `getSolution` method [here](https://github.com/overblog/GraphQLBundle/blob/master/src/Definition/Builder/SchemaBuilder.php#L67). This will throw an exception with a clearer message. Or we could also just check types existence on container compile time, so no need to wait on run time to see message? What is the cleaner solution? This fix should maybe be integrated to 0.12 branch so it could be release in next version?

theofidry:
It becomes:

```
Overblog\GraphQLBundle\Resolver\UnresolvableException: Could not found type with alias "Foo". Do you forget to define it?
```

which is good as well. I would however:

- Fix the type /found/find/s
- Still throw an exception: you are supposed to return a type or at the very least an object, not `null`

mcg-web:
This look good to me, do you want to make the changes yourself ?

theofidry:
If that can wait a week yes :)

theofidry:
The failure looks unrealted

mcg-web:
It's ok, failure is accepted for Symfony 4.4 dev and 5.0 dev.

src/Resolver/AbstractResolver.php
```diff
@@ -88,7 +90,7 @@ protected function onLoadSolution($solution): void
     /**
      * @param string $id
      *
-     * @return mixed
+     * @return Type|null
```
mcg-web: This interface is also used for resolvers and mutations resolver.

theofidry:
Ah I see; well you can revert that with the editor then I allowed the
edits; I don’t have access to my computer for a while :P

On Fri 21 Jun 2019 at 21:07, Jeremiah VALERIE <notifications@github.com>
wrote

mcg-web:
I apply the changes, thank you for feedback an PR.

PR #510, opened 2019-06-27
Validate array of input argument types by ArgumentsTransformer.

andreyserdjuk:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets |
| License       | MIT

### DIFF
--- a/src/DependencyInjection/Compiler/ConfigParserPass.php
+++ b/src/DependencyInjection/Compiler/ConfigParserPass.php
@@ -2,22 +2,24 @@
 
 declare(strict_types=1);
 
-namespace Overblog\GraphQLBundle\DependencyInjection;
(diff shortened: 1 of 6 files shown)


andreyserdjuk:
Error "TypeError: abs() expects parameter 1 to be int or float, string given" - because of last nigthly PHP build abs() asserts input to be numeric value.
As Symfony yaml parser causes this error we can only fix Symfony component or not to test with PHP nightly build for now.

This error caused by changing Symfony 4.4 -> 5.0 !tagged syntax:
"!tagged" is deprecated since Symfony 4.4 and will be removed in 5.0, use "!tagged_iterator" instead.

PR #511, opened 2019-06-28
Compiler pass instead of !tagged expression, SF 5 tests run fix

andreyserdjuk:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | no
| Documented?   | no
| Fixed tickets | #... 
| License       | MIT

Because since SF 4.4 !tagged expression should be renamed to !tagged_iterator the only way to keep bundle compatible is not to use both of these keywords.

Also fixed tests run for SF 5.0.* - where Translator component is not included by default.

I propose to not to test with php nightly builds. Because even if we can adopt this repo for php8 - Symfony code causes unit-tests errors. Because .travis matrix in Symfony does not include php nightly yet. I really doubt that Symfony community will approve my fix for php that even not included in .travis.

### DIFF
--- a/src/DependencyInjection/Compiler/ConfigProcessorPass.php
+++ b/src/DependencyInjection/Compiler/ConfigProcessorPass.php
@@ -1,0 +1,31 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Overblog\GraphQLBundle\DependencyInjection\Compiler;
+
+use Overblog\GraphQLBundle\Definition\ConfigProcessor;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+class ConfigProcessorPass implements CompilerPassInterface
+{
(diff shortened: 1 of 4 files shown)


src/DependencyInjection/Compiler/ConfigProcessorPass.php
```diff
@@ -0,0 +1,31 @@
...
+use Overblog\GraphQLBundle\Definition\ConfigProcessor;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+class ConfigProcessorPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container): void
+    {
+        if (!$container->has(ConfigProcessor::class)) {
```
mcg-web: Remove this please, this is useless condition the bundle always has this service.

andreyserdjuk: Of course!

PR #516, opened 2019-07-03
Validate every input object from arguments array

andreyserdjuk:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes (not by myself)
| License       | MIT

This feature applies validator for each element in arguments array.  
For example: [OurObjectWithConstraints!]
In yaml it looks like:
```yaml
resolve: "@=resolver('some_resolver', arguments({filter: '[OurObjectWithConstraints!]'}, args))"
```


mcg-web:
Thank you @andreyserdjuk ! I rebased this new feature to master since 0.12 is now feature freeze. @Vincz  will do the review then we will merge this if it is ok for him too.

andreyserdjuk:
> Maybe move the errors inside the check of the validator, instantiate it as a `ConstraintViolationList` instead of an array, and populate it with the errors with the `addAll`.
> Otherwise, looks good to me :)

Very good, I'm agree and also thought about pros and cons:
+ code is more clear
- ArgumentsTransformer will be aware of exact ConstraintViolationListInterface implementation. I mean if some hypothetical future Symfony version will use another implementation of ConstraintViolationListInterface.

mcg-web:
Thank you @Vincz and @andreyserdjuk !

PR #536, opened 2019-07-23
feature/validation

murtukov:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no (maybe?)
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes

### Documentation:
Please first read the [documentation](https://github.com/murtukov/GraphQLBundle/blob/feature/validation_v.0.13/docs/validation/index.md) completely. If some sections are not clear let me know, I'll try to rewrite them in more clear words.

Think about the titles, maybe some of them should be changed. There is a section called **How does it work?**. Is the name ok or should we rename it to something like **Internals**? Any other suggestions?

### Architechture:
When you call `graphql:compile`, the Generator creates definition classes. Currently all resolvers always look something like this:
```php
'resolve' => function ($value, $args, $context, ResolveInfo $info) use ($globalVariable) {
    return $globalVariable->get('mutationResolver')->resolve(["createUser", [0 => $args]]);
},
```
there is no extra code, just calling `$globalVariable->get('...')->resolve([...])`.

I modified the generator classes so, that when you add validation to your configuration and call `graphql:compile`, the Generator creates an extra code in the resolver of your definition classes, something like this:

```php
<?php
// ...
'resolve' => function ($value, $args, $context, ResolveInfo $info) use ($globalVariable) {
     $validator = new InputValidator(
         func_get_args(),
         $globalVariable->get('container')->get('validator'),
         $globalVariable->get('validatorFactory'),
         [
             'class' => [
                 'link' => null,
                 'constraints' => [
                     new Assert\Expression('arguments([], [])'),
                 ],
                 'cascade' => null
             ],
             'properties' => [
                 'username' => [
                     'link' => null,
                     'constraints' => [
                         new Assert\Length([
                             'min' => 6, 
                             'max' => 32, 
                             'minMessage' => 'createUser.username.length.min'
                         ]),
                         new Assert\NotBlank([
                             'message' => 'Something is empty'
                         ]),
                     ],
                     'cascade' => null
                 ],
                 'password' => [
                     'link' => null,
                     'constraints' => [
                         new Assert\Length([
                             'min' => 6, 
                             'max' => 32
                         ]),
                         new Assert\IdenticalTo([
                             'propertyPath' => 'passwordRepeat', 
                             'message' => 'Passwords are not equal!'
                         ]),
                     ],
                     'cascade' => null
                 ],
                 'passwordRepeat' => null,
                 'birthday' => [
                     'link' => null,
                     'constraints' => [
                         new Assert\Date(),
                         new Assert\Callback([
                             'callback' => [
                                 'App\Util\Validator', 
                                 'greaterThan'
                             ], 
                             'payload' => '-18 years'
                         ]),
                     ],
                     'cascade' => null
                 ],
             ]
      ]
     );

     return $globalVariable->get('mutationResolver')->resolve(["createUser", [0 => $args, 1 => $validator]]);
    },
//...
```
This will be done by the `generateExtraCode()` method declared in the `lib\` folder as `abstract`, so that the code under the `lib\` folder remains generic and it's implementation locates under the `src\` directory.

The main reason why I decided to make it all by Generator is because of its performance: 
1) The `InputValidator` class will be instantiated only when you call a resolver.
2) If you have for example 20 resolvers, it won't instantiate all validators, but only those, which are in the requested resolvers.

So it doesn't create any services and therefore creates no overhead.

The generator classes won't generate extra code if you dont declare the `validator` variable in the expression e.g.:
```yaml
# This will trigger the generator to generate extra code in 
# the resolver callback
resolve: "@=res('my_resolver', [args, validator])"

# and this won't generate anything in the resolver callback
resolve: "@=res('my_resolver', [args])"
```

if you have any thoughts about this solution, let me know.

### Formatting of validation errors
It uses the built in feature [Error handling](https://github.com/murtukov/GraphQLBundle/blob/master/docs/error-handling/index.md) to return validation errors to the client. I created a class `Overblog\GraphQLBundle\Validator\Formatter` to change the response for exceptions of class `ArgumentsValidationException`. 

### Validators
All the validation logic is located under the `src\Validator` directory. 

### Annotations and `ArgumentsTransformer` class.
@Vincz I assume that the annotation feature is mainly your work. Currently my validator implementation works only for yaml configuration. I would like to also add support for annotations, but first I would like to discuss your implementation of the class `ArgumentsTransformer` with you , because it has it's own validation mechanism which conflicts with my implementation. I will create a separate issue for that, but for now I ask you to stop any further development of the `ArgumentsTransformer`. Maybe it will be removed or modified, to use this validation feature (depends on our discussion).

___

I will edit and add new stuff to this comment during the discussion.


lib/generator/src/Generator/AbstractTypeGenerator.php
```diff
@@ -29,7 +31,7 @@ abstract class AbstractTypeGenerator extends AbstractClassGenerator
```
murtukov: Addes a new placeholder `%s`. This is where the extra code will be inserted

lib/generator/src/Generator/AbstractTypeGenerator.php
```diff
@@ -67,6 +69,11 @@ function (%s) <closureUseStatements>{
      */
     protected $cacheDirMask;
```
murtukov: The `$currentlyGeneratedClass` is used to know in any generator class which definition is being generated right now.

lib/generator/src/Generator/AbstractTypeGenerator.php
```diff
@@ -29,7 +31,7 @@ abstract class AbstractTypeGenerator extends AbstractClassGenerator
 
     protected const CLOSURE_TEMPLATE = <<<EOF
 function (%s) <closureUseStatements>{
-<spaces><spaces>return %s;
+<spaces><spaces>%sreturn %s;
```
murtukov: Added a new placeholder `%s`. This is where the extra code will be inserted.

lib/generator/src/Generator/AbstractTypeGenerator.php
```diff
@@ -67,6 +69,11 @@ function (%s) <closureUseStatements>{
      */
     protected $cacheDirMask;
 
+    /**
+     * @var string
+     */
+    protected $currentlyGeneratedClass;
```
murtukov: The `$currentlyGeneratedClass` is used to know in any generator class which definition is being generated right now.

lib/generator/src/Generator/AbstractTypeGenerator.php
```diff
@@ -226,15 +233,19 @@ protected function callableCallbackFromArrayValue(array $value, string $key, ?st
                     $compilerNames = $matches[1] ?? [];
                 }
             }
+
+            $extraCode = $this->generateExtraCode($value, $key, $argDefinitions, $default, $compilerNames);
```
murtukov: This is where the validation code will be generated and inserted in the resolver callback.

lib/generator/src/Generator/AbstractTypeGenerator.php
```diff
@@ -373,6 +386,20 @@ public function generateClass(array $config, ?string $outputDirectory, int $mode
...
+     * Adds an extra code into resolver closure before 'return' statement
+     *
+     * @param array         $value
+     * @param string        $key
+     * @param string|null   $argDefinitions
+     * @param string        $default
+     * @param array|null    $compilerNames
+     * @return string|null
+     */
+    abstract protected function generateExtraCode(array $value, string $key, ?string $argDefinitions = null, string $default = 'null', array &$compilerNames = null): ?string;
```
murtukov: An abstract method to insert an extra code into resolver closures.

src/ExpressionLanguage/ExpressionFunction/GraphQL/Relay/IdFetcherCallback.php
```diff
@@ -18,6 +19,19 @@ function ($idFetcher) {
                 $code .= 'return '.$idFetcher.'; }';
 
                 return $code;
+            },
```
murtukov: This is a new non-documented expression functions which needs an evaluator. Currently it has only a dummy evaluator, which throws an exception, if used in the `Expression` validation constraint.

src/ExpressionLanguage/ExpressionFunction/GraphQL/Relay/MutateAndGetPayloadCallback.php
```diff
@@ -18,6 +19,19 @@ function ($mutateAndGetPayload) {
                 $code .= 'return '.$mutateAndGetPayload.'; }';
 
                 return $code;
+            },
```
murtukov: This is a new non-documented expression functions which needs an evaluator. Currently it has only a dummy evaluator, which throws an exception, if used in the `Expression` validation constraint.

src/ExpressionLanguage/ExpressionFunction/GraphQL/Relay/ResolveSingleInputCallback.php
```diff
@@ -18,6 +19,19 @@ function ($resolveSingleInput) {
                 $code .= 'return '.$resolveSingleInput.'; }';
 
                 return $code;
+            },
```
murtukov: This is a new non-documented expression functions which needs an evaluator. Currently it has only a dummy evaluator, which throws an exception, if used in the `Expression` validation constraint.

akomm:
@murtukov validator injection is good. Also the option not to throw. I had similar feature in mind for the `ArgumentsTransformer`. 

Reason: There are few opinionated way handling user input errors. Not all of them are handled via `error` and/or `extension`. Currently `ArgumentsTransformer` throws and does not ask questions. I wanted to add a feature to either being able to inject `errors` (ConstraintViolations) or `validator`. Would be kind of obsolete with this PR :)

murtukov:
@akomm 
> Would it be possible to directly inject errors via expression, so ->validate is called implicitly? Not that big of a deal if not.

It's already possible to inject errors from the expression:
```yaml
@=res('myResolver', [args, validator.validate(null, false)])
```
but it doesn't look good. Maybe we should implement something like this:
```yaml
@=res('myResolver', [args, errors])
```

Expression language is a nice feature but its usage should be reduced to minumum, because configuration should be **declarative** and not **imperative**, it should'n contain logic. 
For example instead of this:
`@=res('myResolver', [args['name'], args['surname'], args['email'], args['password']])`
we should do this: 
`@=res('myResolver', [args])`
and destructurize the `args` variable in the php.

I believe that we are abusing the expression language. I am currently thinking about solutions to reduce the usage of expression language in the definition configs without having any drawbacks, but it's a topic of a separate _issue_.

#### About the `ArgumentsTransformer`:
This class and the feature should be reconsidered and reworked as it has many problems. **Transformation** and **Validation** are 2 different operations and should be decoupled. I don't know, why the ArgumentsTransformer makes validation. We should have a transformer and a validator as 2 completely separated features.
Secondly the transformation configuration shouldn't be done in the expression (as it currently is). I will create a separate issue for this.

murtukov:
@akomm Btw what do you mean with "Would be kind of obsolete with this PR"? Didn't understand that

akomm:
@murtukov yes, I know I can just invoke it, but exactly for the reason you explained I did propose `errors`.

The feature I wanted to add to `ArgumentsTransforber` would be obsolete with your PR, that is what I meant :)

murtukov:
@akomm The problem is, that this validation feature doesn't currently work with annotations. I will fix it in future, but first we need to get rid of it's old validation mechanism

akomm:
I was using `yaml` & `ResolverInterface`, `MutationInterface`, `AliasedInterface` all the time. I was just testing annotation as option for several reasons and encountered some issues. 

I agree validation and hydration should be separate things. However then validation maybe should not alter data and make objects then, otherwise if you add layer for hydration and validation is optional, the data received would be different with or without validation.

murtukov:
@akomm I don't understand, what problem do you mean? My implementation does NOT alter the input data. You will receive a raw data in your resolver.

akomm:
> @akomm I don't understand, what problem do you mean? My implementation does NOT alter the input data. You will receive a raw data in your resolver.

You are right. Its okay.

I think if you miss the `temporary` word [here](https://github.com/murtukov/GraphQLBundle/blob/feature/validation_v.0.13/docs/validation/index.md#how-does-it-work) you get the wrong impression of it being a permanent data transformation continuing the docs.

murtukov:
@akomm thanx for the feedback, I'll add a note to the doc, that the validation objects are temporar and arguments won't be transformed.

lib/generator/src/Generator/AbstractTypeGenerator.php
mcg-web: this comment should be add to phpdoc directly.

src/Exception/EvaluatorIsNotAllowedException.php
```diff
@@ -0,0 +1,19 @@
...
+
+declare(strict_types=1);
+
+namespace Overblog\GraphQLBundle\Exception;
+
+use Throwable;
+
+class EvaluatorIsNotAllowedException extends \Exception
+{
+    public function __construct($expressionFunctionName, $code = 0, Throwable $previous = null)
```
mcg-web:
```suggestion
    public function __construct(string $expressionFunctionName, int $code = 0, Throwable $previous = null)
```

src/Exception/EvaluatorIsNotAllowedException.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Overblog\GraphQLBundle\Exception;
```
mcg-web: this should be moved in ExpressionLanguage namespace

src/ExpressionLanguage/ExpressionFunction/GraphQL/Mutation.php
```diff
@@ -13,7 +14,11 @@ public function __construct($name = 'mutation')
         parent::__construct(
             $name,
             function ($alias, $args = '[]') {
-                return \sprintf('$globalVariable->get(\'mutationResolver\')->resolve([%s, %s])', $alias, $args);
+                return "\$globalVariable->get('mutationResolver')->resolve([$alias, $args])";
+            },
+            // This expression function is not designed to be used by it's evaluator
+            function (): void {
```
mcg-web:
```suggestion
            function () us ($name): void {
```

src/ExpressionLanguage/ExpressionFunction/GraphQL/Mutation.php
```diff
@@ -13,7 +14,11 @@ public function __construct($name = 'mutation')
...
+                throw new EvaluatorIsNotAllowedException('resolveSingleInputCallback');
```
mcg-web:
```suggestion
                throw new EvaluatorIsNotAllowedException($name);
```

src/ExpressionLanguage/ExpressionFunction/GraphQL/Relay/FromGlobalID.php
```diff
@@ -5,19 +5,19 @@
 namespace Overblog\GraphQLBundle\ExpressionLanguage\ExpressionFunction\GraphQL\Relay;
 
 use Overblog\GraphQLBundle\ExpressionLanguage\ExpressionFunction;
+use Overblog\GraphQLBundle\Relay\Node\GlobalId;
```
mcg-web: Scrutinizer: This use statement conflicts with another class in this namespace, ``Overblog\GraphQLBundle\ExpressionLanguage\ExpressionFunction\GraphQL\Relay\GlobalId``. Consider defining an alias.

src/ExpressionLanguage/ExpressionFunction/GraphQL/Relay/GlobalID.php
```diff
@@ -5,23 +5,29 @@
...
-                return \sprintf(
-                    '\%s::toGlobalId(%s, %s)',
-                    \Overblog\GraphQLBundle\Relay\Node\GlobalId::class,
-                    \sprintf($typeNameEmpty ? '$info->parentType->name' : '%s', $typeName),
-                    $id
-                );
+                return \sprintf('\%s::toGlobalId(%s, %s)', GlobalIdHelper::class, $typeName, $id);
+            },
+            function ($arguments, $id, $typeName = null) {
+                $typeName = $this->isTypeNameEmpty($typeName) ? $arguments['info']->parentType->name : $typeName;
```
mcg-web:
```suggestion
                $typeName = empty($typeName) ? $arguments['info']->parentType->name : $typeName;
```

src/ExpressionLanguage/ExpressionFunction/GraphQL/Relay/GlobalID.php
```diff
@@ -5,23 +5,29 @@
...
+
+                return GlobalIdHelper::toGlobalId($typeName, $id);
             }
         );
     }
+
+    private function isTypeNameEmpty($typeName)
```
mcg-web:
```suggestion
    private function isTypeNameEmpty($typeName): bool
```

src/ExpressionLanguage/ExpressionFunction/GraphQL/IsTypeOf.php
```diff
@@ -8,12 +8,15 @@
...
+    public function __construct()
     {
         parent::__construct(
-            $name,
+            'isTypeOf',
             function ($className) {
                 return \sprintf('(($className = %s) && $value instanceof $className)', $className);
+            },
+            function ($arguments, $className): bool {
+                return $className && $arguments['prevValue'] instanceof $className;
```
mcg-web: `prevValue` ?

src/Exception/EvaluatorIsNotAllowedException.php
```diff
@@ -0,0 +1,19 @@
...
+{
+    public function __construct($expressionFunctionName, $code = 0, Throwable $previous = null)
+    {
+        parent::__construct(
+            "The expression function '$expressionFunctionName' cannot be used by it's evaluator.",
+            $code,
+            $previous
+        );
+    }
+}
```
mcg-web:
```suggestion
    public function __invoke(): void
    {
        throw $this;
    }
}
```

src/ExpressionLanguage/ExpressionFunction/GraphQL/Relay/IdFetcherCallback.php
```diff
@@ -18,6 +19,10 @@ function ($idFetcher) {
...
+            // This expression function is not designed to be used by it's evaluator
+            function (): void {
```
mcg-web: you can use `new EvaluatorIsNotAllowedException($name)` since Exception is callable

src/ExpressionLanguage/ExpressionFunction/GraphQL/Relay/MutateAndGetPayloadCallback.php
```diff
@@ -18,6 +19,10 @@ function ($mutateAndGetPayload) {
...
+            // This expression function is not designed to be used by it's evaluator
+            function (): void {
```
mcg-web: you can use `new EvaluatorIsNotAllowedException($name)` since Exception is callable

src/ExpressionLanguage/ExpressionFunction/GraphQL/Resolver.php
```diff
@@ -12,8 +13,12 @@ public function __construct($name = 'resolver')
     {
         parent::__construct(
             $name,
-            function ($alias, $args = '[]') {
-                return \sprintf('$globalVariable->get(\'resolverResolver\')->resolve([%s, %s])', $alias, $args);
+            function (string $alias, string $args = '[]'): string {
+                return "\$globalVariable->get('resolverResolver')->resolve([$alias, $args])";
+            },
+            // This expression function is not designed to be used by it's evaluator
+            function (): void {
```
mcg-web: you can use `new EvaluatorIsNotAllowedException($name)` since Exception is callable

src/ExpressionLanguage/ExpressionFunction/GraphQL/Relay/ResolveSingleInputCallback.php
```diff
@@ -18,6 +19,10 @@ function ($resolveSingleInput) {
...
+            // This expression function is not designed to be used by it's evaluator
+            function (): void {
```
mcg-web: you can use `new EvaluatorIsNotAllowedException($name)` since Exception is callable

src/ExpressionLanguage/ExpressionFunction/Security/HasAnyPermission.php
```diff
@@ -6,17 +6,25 @@
 
 use Overblog\GraphQLBundle\ExpressionLanguage\ExpressionFunction;
 use Overblog\GraphQLBundle\Generator\TypeGenerator;
+use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;
 
 final class HasAnyPermission extends ExpressionFunction
 {
-    public function __construct($name = 'hasAnyPermission')
+    public function __construct(AuthorizationCheckerInterface $authorizationChecker, $name = 'hasAnyPermission')
```
mcg-web: why keeping `$name` argument sometime?

src/ExpressionLanguage/ExpressionFunction/GraphQL/IsTypeOf.php
murtukov: As you know there are predefined expression variables used in resolve context: `args`, `context`, `info` and `value`.

The [Expression](https://symfony.com/doc/current/reference/constraints/Expression.html) validation constraint also has a predefined variable called `value`.

The thing is, that you can use all variables from the resolve context in the expression context, so the variable names conflict. Thats why, when you want to access the variable `value` from the resolve context in your expression, you need to use `prevValue` instead, because the Expression constraint has it's own `value`.

Please reference the [documentation](https://github.com/murtukov/GraphQLBundle/blob/feature/validation_v.0.13/docs/validation/index.md#using-built-in-expression-functions).

src/Exception/EvaluatorIsNotAllowedException.php
murtukov: Will do it.
I guess the `ArgumentsValidationException` class should be moved to Validator namespace too?

src/ExpressionLanguage/ExpressionFunction/GraphQL/Relay/GlobalID.php
murtukov: Why replaced with `empty`? I just used the code from the compiler callback

src/ExpressionLanguage/ExpressionFunction/GraphQL/Relay/FromGlobalID.php
murtukov: This code must be old, because I already have this line using an alias:
```
use Overblog\GraphQLBundle\Relay\Node\GlobalId as GlobalIdHelper;
```

Strange.

src/ExpressionLanguage/ExpressionFunction/Security/HasAnyPermission.php
murtukov: I initially planned to keep the `$name` variable only for classes with alias (`mutation`, `resolver`, `parameter` and `service`). I just forgot to remove them from other classes.

Fixed that.

src/ExpressionLanguage/ExpressionFunction/GraphQL/Relay/GlobalID.php
mcg-web: compiler callback is for code string, evaluator is for php values (`boolean`, `string`, ...) so `empty` does the because it also cover `false` vs `isTypeNameEmpty` method that cover `'false'`.

src/Exception/EvaluatorIsNotAllowedException.php
mcg-web: yes an Exception dir should be create under each related Namespace if possible

src/ExpressionLanguage/ExpressionFunction/GraphQL/Relay/FromGlobalID.php
mcg-web: I don't see alias on [your branch](https://github.com/murtukov/GraphQLBundle/blob/feature/validation_v.0.13/src/ExpressionLanguage/ExpressionFunction/GraphQL/Relay/FromGlobalID.php#L8), maybe you loose the change on a rebase?

src/Validator/Constraints/ExpressionValidator.php
```diff
@@ -0,0 +1,66 @@
...
+                ->setParameter('{{ value }}', $this->formatValue($value, self::OBJECT_TO_STRING))
+                ->setCode(Expression::EXPRESSION_FAILED_ERROR)
+                ->addViolation();
+        }
+    }
+
+    private function getExpressionLanguage()
+    {
+        if (null === $this->expressionLanguage) {
+            if (!\class_exists('Symfony\Component\ExpressionLanguage\ExpressionLanguage')) {
```
mcg-web: this condition is useless since expression language is a requirement of the bundle

src/Validator/Constraints/ExpressionValidator.php
```diff
@@ -0,0 +1,66 @@
...
+                throw new LogicException('Unable to use expressions as the Symfony ExpressionLanguage component is not installed.');
+            }
+            $this->expressionLanguage = new ExpressionLanguage();
```
mcg-web: why not injecting the bundle expression language service?

src/Validator/Constraints/ExpressionValidator.php
```diff
@@ -0,0 +1,66 @@
...
+
+        $variables = $constraint->values;
+        $variables['value'] = $value;
+
+        $object = $this->context->getObject();
+
+        $variables['this'] = $object;
+
+        if ($object instanceof ValidationNode) {
+            $variables['prevValue'] = $object->getResolverArg('value');
```
mcg-web: `rawValue` instead of  `prevValue`?

src/Validator/Formatter.php
```diff
@@ -0,0 +1,32 @@
...
+declare(strict_types=1);
+
+namespace Overblog\GraphQLBundle\Validator;
+
+use Overblog\GraphQLBundle\Event\ErrorFormattingEvent;
+use Overblog\GraphQLBundle\Validator\Exception\ArgumentsValidationException;
+
+class Formatter
+{
+    public function onErrorFormatting(ErrorFormattingEvent $event): void
```
mcg-web: Can you add a comment with an example of the output of errors after formatting please?

src/Validator/Formatter.php
```diff
@@ -0,0 +1,32 @@
...
+            $code = [];
+
+            $violations = $error->getViolations();
+            foreach ($violations as $violation) {
+                $state[$violation->getPropertyPath()][] = $violation->getMessage();
+                $code[$violation->getPropertyPath()][] = $violation->getCode();
+            }
+
+            $formattedError = $event->getFormattedError();
+            $formattedError->offsetSet('state', $state);
```
mcg-web: error should contain only `message`, `locations`, `path` and `extensions` as root entry  all additional entry should be add under `extensions` (see [the specs](https://graphql.github.io/graphql-spec/June2018/#sec-Errors))

src/Validator/InputValidator.php
```diff
@@ -0,0 +1,281 @@
...
+            }
+
+            foreach ($params ?? [] as $key => $value) {
+                if (null === $value) {
+                    continue;
+                }
+
+                switch ($key) {
+                    case 'link':
+                        [$FQCN, $property, $type] = $value;
```
mcg-web: use lower case for var `$FQCN` please

src/Validator/InputValidator.php
```diff
@@ -0,0 +1,281 @@
...
+        foreach ($values as $value) {
+            $collection[] = $this->createObjectNode($value, $type, $parent);
+        }
+
+        return $collection;
+    }
+
+    /**
+     * @param array                           $value
+     * @param ObjectType|InputObjectType|Type $type
```
mcg-web:
```suggestion
     * @param ObjectType|InputObjectType $type
```

src/Validator/Mapping/MetadataFactory.php
```diff
@@ -0,0 +1,42 @@
...
+    public function getMetadataFor($object)
+    {
+        if ($object instanceof ValidationNode) {
+            return $this->metadataPool[$object->getName()];
+        }
+
+        throw new NoSuchMetadataException();
+    }
+
+    public function hasMetadataFor($object)
```
mcg-web:
```suggestion
    public function hasMetadataFor($object): bool
```

src/Validator/Mapping/MetadataFactory.php
```diff
@@ -0,0 +1,42 @@
...
+class MetadataFactory implements MetadataFactoryInterface
+{
+    private $metadataPool;
+
+    public function __construct()
+    {
+        $this->metadataPool = [];
+    }
+
+    public function getMetadataFor($object)
```
mcg-web: add ouput type please

src/Validator/ValidationNode.php
```diff
@@ -0,0 +1,152 @@
...
+ * instantiated.
+ */
+class ValidationNode
+{
+    private const KNOWN_VAR_NAMES = ['value', 'args', 'context', 'info'];
+
+    /**
+     * @var ObjectType|InputObjectType|Type
+     */
+    private $__type;
```
mcg-web: why the prefix `__` for vars name?

src/Validator/Formatter.php
murtukov: Comment to the pho or just here?

src/Validator/ValidationNode.php
murtukov: The objects of the class ValidationNode are populated with user input data to be validated later. The properties are created dynamically from argument names of GraphQL types. The double underscore prefix should prevent name conflicts, if a user uses in his types such names as ‘type’ or ‘fieldName’ etc.

src/Validator/Constraints/ExpressionValidator.php
murtukov: No, `rawValue` would rather mean an "unprepared" or an "unprocessed" value. But the "prevValue" represents the value of the "previous" resolver or "parent" resolver. So the names should be `prevValue` or `parentValue`.

src/Validator/Constraints/ExpressionValidator.php
murtukov: Fixed

src/Validator/Constraints/ExpressionValidator.php
murtukov: This class is not declared in the bundle's `services.yml`. The declaration of this class as a service is fully on the users and is documented [here](https://github.com/murtukov/GraphQLBundle/blob/feature/validation_v.0.13/docs/validation/index.md#using-built-in-expression-functions).

This class contained this weird logic, because it's was mostly a copy-paste of the original class from the Symfony. Refactored.

src/Validator/Formatter.php
murtukov: Will be refactored according to the specification.

src/Validator/InputValidator.php
murtukov: Fixed

src/Validator/InputValidator.php
murtukov: Fixed

src/Validator/Mapping/MetadataFactory.php
murtukov: Fixed

src/Validator/Constraints/ExpressionValidator.php
mcg-web: ok for `parentValue`

src/Validator/Formatter.php
mcg-web: to php please

src/Validator/Formatter.php
murtukov: Done

src/Validator/Formatter.php
murtukov: I added a [section](https://github.com/murtukov/GraphQLBundle/blob/feature/validation_v.0.13/docs/validation/index.md#error-messages) to the documentation and added a @link annotation to the php, since the inline example would be too large. 
Also response examples could be found under the path: `GraphQLBundle\tests\Functional\Validator\ExpectedErrors.php`.

src/Validator/Constraints/ExpressionValidator.php
murtukov: Done!

src/ExpressionLanguage/ExpressionFunction/GraphQL/Relay/ResolveSingleInputCallback.php
murtukov: Done!

src/ExpressionLanguage/ExpressionFunction/GraphQL/Resolver.php
murtukov: Done!

src/ExpressionLanguage/ExpressionFunction/GraphQL/Relay/MutateAndGetPayloadCallback.php
murtukov: Done!

src/ExpressionLanguage/ExpressionFunction/GraphQL/Relay/IdFetcherCallback.php
murtukov: Done!

src/ExpressionLanguage/ExpressionFunction/GraphQL/Relay/GlobalID.php
murtukov: Fixed!

composer.json
```diff
@@ -36,8 +36,10 @@
         "symfony/event-dispatcher": "^4.3 || ~5.0",
         "symfony/expression-language": "^4.3 || ~5.0",
         "symfony/framework-bundle": "^4.3 || ~5.0",
+        "symfony/http-kernel": "^4.3",
         "symfony/options-resolver": "^4.3 || ~5.0",
         "symfony/property-access": "^4.3 || ~5.0",
+        "symfony/translation": "^4.3",
```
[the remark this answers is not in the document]
murtukov: Moved to the `suggestions` section.

composer.json
```diff
@@ -36,8 +36,10 @@
         "symfony/event-dispatcher": "^4.3 || ~5.0",
         "symfony/expression-language": "^4.3 || ~5.0",
         "symfony/framework-bundle": "^4.3 || ~5.0",
+        "symfony/http-kernel": "^4.3",
```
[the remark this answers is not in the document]
murtukov: Removed the line.

Issue #561, opened 2019-09-26
Sort fields and arguments


ooflorent:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? |  no
| RFC?             | no
| Version/Branch   | n/a

Fields and arguments could be sorted to improve the type readability.

murtukov:
@ooflorent You mean yaml files?

akomm:
@ooflorent is think rather introspection and dump-output (.graphql) ?

ooflorent:
Okay, the idea is this:

```yaml
MyType:
  type: object
  config:
    fields:
      b:
        type: String
      a:
        type: String
```

Should produce this:

```php
new ObjectType([
  'fields' => [
    'a' => …
    'b' => …
  ]
])
```

Instead of this:

```php
new ObjectType([
  'fields' => [
    'b' => …
    'a' => …
  ]
])
```

By having the fields (and arguments) sorted in the type definition, they will be sorted everywhere.

PR #562, opened 2019-09-27
Sort fields, args and values in generated types

mcg-web:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | #561 
| License       | MIT

### DIFF
--- a/lib/generator/src/Generator/AbstractTypeGenerator.php
+++ b/lib/generator/src/Generator/AbstractTypeGenerator.php
@@ -74,15 +74,20 @@ EOF;
      */
     protected $currentlyGeneratedClass;
 
+    /** @var bool */
+    protected $sort;
+
     /**
      * @param string $classNamespace The namespace to use for the classes.
      * @param string[]|string $skeletonDirs
      * @param int $cacheDirMask
+     * @param bool $sort
      */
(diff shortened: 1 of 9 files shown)


murtukov:
@mcg-web @ooflorent 

I definitely disagree, that the fields should be sorted. 
If I declared fields in a certain way, then I expect the generated fields be sorted in the same way. 

Why should they automatically be sorted?

ooflorent:
> This is subjective.

Absolutely. However currently it is impossible to keep them sorted when using decorators.
Maybe the feature could be controlled by flag.

What are your thoughts about it @mcg-web?

murtukov:
@ooflorent It can easily be made optional. I would put it under:

```yaml
overblog_graphql:
    compiler:
        sort_fields: true
```

mcg-web:
I'll add a flag cause it's true that having PK like id first can make thing easier. Further more if your object has a lot of field. But what the best default value for this flag? :thinking:

ooflorent:
It's better to default to `false` to avoid breaking changes if people rely on property order.

murtukov:
This PR is outdated. It should be created again on the new codebase. The newly integrated generator library supports sorting array keys in ASC and DESC orders. Feel free to open an issue @ooflorent

PR #564, opened 2019-10-02
Fix @validator.validator_factory optional service

mcg-web:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Documented?   | no
| Fixed tickets | none
| License       | MIT

the service `validator.validator_factory` comes with validator component so should not be required since validator itself is optional

```
The service "Overblog\GraphQLBundle\Definition\Builder\TypeFactory" has a dependency on a non-existent service "validator.validator_factory".
```

### DIFF
--- a/src/Resources/config/services.yaml
+++ b/src/Resources/config/services.yaml
@@ -98,12 +98,3 @@ services:
     Overblog\GraphQLBundle\Validator\Formatter:
         tags:
             - { name: kernel.event_listener, event: graphql.error_formatting, method: onErrorFormatting }
-
-    Overblog\GraphQLBundle\Validator\ValidatorFactory:
-        public: false
-        arguments:
-            - '@?validator'
-            - '@validator.validator_factory'
-            - '@?translator.default'
-        tags:
-            - { name: overblog_graphql.global_variable, alias: validatorFactory, public: false }
--- a/src/Validator/ValidatorFactory.php
+++ b/src/Validator/ValidatorFactory.php
@@ -12,13 +12,11 @@ use Symfony\Contracts\Translation\TranslatorInterface;
 
 class ValidatorFactory
 {
-    private $defaultValidator;
     private $defaultTranslator;
     private $constraintValidatorFactory;
 
-    public function __construct(?ValidatorInterface $validator, ConstraintValidatorFactoryInterface $constraintValidatorFactory, ?TranslatorInterface $translator)
+    public function __construct(ConstraintValidatorFactoryInterface $constraintValidatorFactory, ?TranslatorInterface $translator)
     {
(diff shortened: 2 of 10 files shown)


src/Validator/ValidatorFactory.php
```diff
@@ -27,9 +27,12 @@ public function createValidator(MetadataFactory $metadataFactory): ValidatorInte
     {
         $builder = Validation::createValidatorBuilder()
             ->setMetadataFactory($metadataFactory)
-            ->setConstraintValidatorFactory($this->constraintValidatorFactory)
         ;
 
+        if (null !== $this->constraintValidatorFactory) {
+            $builder->setConstraintValidatorFactory($this->constraintValidatorFactory);
+        }
```
murtukov: This block doesn't make much sense, because this class is useless without the `ConstraintValidatorFactory` being injected into it. I would rather add an exception into constructor, as I did in the `InputValidator` class, to prevent the use of this class.

In the case on the `$defaultTranslator` it's perfectly fine if no Translator was injected, because the `InputValidator` can work without it. But without the `ConstraintValidatorFactory` it won't work correctly.

src/Resources/config/services.yaml
```diff
@@ -103,7 +103,7 @@ services:
         public: false
         arguments:
             - '@?validator'
-            - '@validator.validator_factory'
+            - '@?validator.validator_factory'
```
murtukov: I agree with that, it should be nullable

src/Validator/ValidatorFactory.php
```diff
@@ -16,7 +16,7 @@ class ValidatorFactory
     private $defaultTranslator;
     private $constraintValidatorFactory;
 
-    public function __construct(?ValidatorInterface $validator, ConstraintValidatorFactoryInterface $constraintValidatorFactory, ?TranslatorInterface $translator)
+    public function __construct(?ValidatorInterface $validator, ?ConstraintValidatorFactoryInterface $constraintValidatorFactory, ?TranslatorInterface $translator)
     {
         $this->defaultValidator = $validator;
```
mcg-web: @murtukov why this is injected it is never used here?

murtukov: @mcg-web I think I just forgot about it. The `InputValidator` uses 2 validators (default and custom). The default one is used to read metadata from classes (required for the `link` feature). Later I wanted to move both of them into ValidatorFactory, thats why I injected the `$defaultValidator` into `ValidatorFactory`, but seems like I forgot about it or it wasn't necessary anymore, I don't remember.

Anyways you can savely remove it from `ValidatorFactory`.

PR #574, opened 2019-10-11
Documentation

murtukov:
Update the documentation (Chapter "Interface")

### DIFF
--- a/docs/definitions/type-system/interface.md
+++ b/docs/definitions/type-system/interface.md
@@ -159,7 +159,7 @@ class MyResolver implements ResolverInterface, AliasedInterface
         // Get an array of Droid objects from DB
         $droids = $this->droidRepository->getAll();
 
-        // We return an array of mixed results. The 'characterType'
+        // We return an array of mixed results. The 'resolveType'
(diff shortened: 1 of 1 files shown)


docs/definitions/type-system/interface.md
```diff
@@ -1,38 +1,253 @@
...
+type Droid implements Character {
+  id: ID!
+  name: String!
+  friends: [Character]
+  appearsIn: [Episode]!
+  primaryFunction: String
+}
+```
+
+We'll show you how the schema above can be implemented with this bundle. There are two main ways to do
```
mcg-web: There are 4 ways to do `xml`, `graphql`, `yaml` and `annotations`. The faster way to implement a simple schema like that one is `graphql` I will document it

mcg-web:
Thank you :+1:

docs/definitions/type-system/interface.md
murtukov: Yes, thats why i wrote "the are two *main* ways".
I think we should state somewhere in the beginning, e.g. "Introduction", that there are 4 ways to define schema.

PR #584, opened 2019-10-16
update phpstan, phpstan-phpunit and added phpstan/extension-installer

mavimo:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| License       | MIT

 - Update phpstan/phpstan
 - Update phpstan/phpstan-phpunit
 - Add phpstan/extension-installer to auto-register extensions

I have only one question, why do you chose to add the package during the composer step instead of adding it on required-dev? There are some blockers or moving in it should be accepted?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -70,9 +70,9 @@
     "scripts": {
         "test": "bin/phpunit --color=always -v --debug",
         "static-analysis": [
-            "@composer req --ansi --dev 'phpstan/phpstan:^0.10.3' 'phpstan/phpstan-phpunit:^0.10.0'",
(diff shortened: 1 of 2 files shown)


mcg-web:
Thank you @mavimo! We removed the package from `required-dev` because it was conflicting with php version requirement.

mavimo:
@mcg-web if you agree I'll try to move it in `composer.json` (maybe we can find some solution that will make it possible 🙈 )

mcg-web:
If you can make it work a PR are welcome :+1:

Issue #587, opened 2019-10-19
Rework all builders


murtukov:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | no
| Version/Branch   | 0.14

@overblog/graphql 

I strongly believe we should rework all builders in the bundle and here I'll try to explain why.

This is just a draft with my thoughts on how builders could be changed. Use examples as a starting point to come to a common conclusion of how field/fields/args builders should be used/changed/enhaced.

# What is wrong with builders?

1) Currently all builders are working with raw arrays and simply return them. Working with PHP arrays is not the best experience, so I suggest we create new builder classes to get rid of using arrays. This would lead to less errors, more readable code, autosuggestions by IDEs + opens perspectives for some other features in the future, for example using one builder inside another. Examples will be shown below a little later.

2) The configuration flow is pretty messy right now. Take a look at the following example:
	```yaml
	User:
	    type: object
	    config:
	        fields:
	            rawId:
	                # Field builder
	                builder: "RawId"
	                builderConfig:
	                      name: photoID
	                      type: String
	                # Args builder
	                argsBuilder:
	                    builder: "Pager"
	                    config:
	                        defaultLimit: 50
	```
	We have `builder`, `builderConfig`, `argsBuilder`, another `builder` one level deeper and `config`, which actually should be also named `builderConfig`. This is not the most readable type configuration and should be also reworked.

3) Lack of flexibility when it comes to the question **Which part of my GraphQL Type do I want to be generated by a builder?** We have 3 types of builders: **args**, **field** and **fields** builders. So what is the difference? Take a look at this picture:

	![enter image description here](https://sun9-41.userapi.com/c857728/v857728717/9e470/ii7qaLN-TQk.jpg)
As you can see the main difference between builders is that they are responsible for different parts of GraphQL types.
Important is that some builders are supersets of other builders, which means, that they can do the same as their subsets, for example we can replace multiple **field** and **args** builders with one **fields** builder, becase **field** and **args** builders are its subsets.
	```
	Fields Builder > Field Builder > Args Builder
	```
	The problem here is, that a user can't use one builder in which he could decide what he wants to be generated. Instead he needs to choose between 3 predefined builders with static limited responsibility (green, blue and red areas on the pic).	Maybe we can combine all 3 builders into 1 (Type Builder)?

#### Current implementation:
The method gets a single argument with all configs and returns an array.

```php
[...]

PR #590, opened 2019-10-22
Added types configuration callback

danut007ro:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| License       | MIT

This will add a callback that will be called at compile time. It can be used to create or extend existing types. See example in docs.

Thank you.


mcg-web:
Thanks @danut007ro for this PR, I need sometime to think about it, we brainstorming about reworking builders #587, this PR seems to be related and add new point of view. This could replace 3 builders and remove the need of a new builder.

danut007ro:
As a matter of fact, I was thinking about separating the current compiler pass in building and processing the configuration steps. This way, a new compiler pass can be injected between those and modify the built configuration, achieving the same result, but in a more symfony-standard way. I was on the point of modifying this pull request, but I will be waiting for your input. Thanks.

Vincz:
An upcoming change to the configuration system will handle configuration as service (the symfony way).

Issue #585, opened 2019-10-17
Add auto validation capability


murtukov:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | no
| Version/Branch   | 0.13

After using of the validation feature in couple of my projects I noticed some repeating, which can and should be automated with an option to switch it off.

Currently to add validation a user have to inject the `validator` variable into the `resolve` expression and have to add validation rules to every arg/field he wishes to validate, e.g.:

```yaml
Mutation:
    type: object
    config: 
        fields:
            createUser:
                type: String!
                resolve: "@=res('my_resolver', [validator, args])"
                args:
                    username:
                        type: String!
                        validation: 
                            - Length: { min: 1, max: 10 }
```
and the resolver:
```php
public function myResolver(InputValidator $validator, ArgumentInterface $args) 
{
    $validator->validate(); # this line is repeated in every resolver

    // ...
}
```

When a user adds validation rules to a GraphQL type this already shows his intention to validate the input data, so there is no need to additionally inject the `validator` variable only to call its `validate()` method. The only exceptions are when users want to validate with specific validation groups or if they want to customize the output of errors:

```php
public function myResolver(InputValidator $validator, ArgumentInterface $args) 
{
    // first argument is a validation group.
    // second argument is a flag to prevent throwing an 
    // error. Instead it will return an error object.
    $errors = $validator->validate('registration', false);

    // ...
}
```

#### So what can we change?
1) Make validation automatic by default. That means the only thing a user have to define is the `validation` key on a GraphQL type.
2) Introduce new expression variable `errors` which will contain all validation errors. It's required if a user don't want to automatically throw validation errors and want to customize the output. The variable is not only for validation errors, but can be used for other future features, for example it could contain transformation errors occured in the Hydrator.
3) Add option to blacklist or whitelist some types from the automatic validation. This would be usefull if there are 50 types and only 48 of them need to be validated autimatically (or vice versa)

First of all we need to introduce a new option into the main config:
```yaml
overblog_graphql:
    # global settings vor validation
    validation:
        auto: true # true by default. Turns on auto validation for all types
```

If a user wants auto validation only for certain types:
```yaml
overblog_graphql:
    validation:
        auto: [User, Post, Book]
```

If a user wants to validate all types, but exclude few of them:
```yaml
overblog_graphql:
    validation:
        auto: 
            exclude: [User, Post, Book]
```

#### Usage:
[...]

PR #593, opened 2019-10-25
Add auto validation

murtukov:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | #585
| License       | MIT


lib/generator/src/Exception/GeneratorException.php
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * (c) Overblog <http://github.com/overblog/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Overblog\GraphQLGenerator\Exception;
+
+class GeneratorException extends \Exception
```
mcg-web:
```suggestion
class GeneratorException extends \RuntimeException
```

src/Generator/TypeGenerator.php
```diff
@@ -273,26 +274,93 @@ protected function generateInputFields(array $config): string
...
+     */
+    protected function generateValidation(array $rules, bool $autoValidation, bool $autoThrow, ?array $groups): string
     {
         $code = $this->processTemplatePlaceHoldersReplacements('ValidatorCode', $rules, self::DEFERRED_PLACEHOLDERS);
         $code = $this->prefixCodeWithSpaces($code, 2);
 
+        if ($autoValidation) {
+            $code .= "\n\n<spaces><spaces>";
+
+            if ($groups) {
```
mcg-web:
```suggestion
            if (null === $groups) {
```

src/Generator/TypeGenerator.php
```diff
@@ -273,26 +274,93 @@ protected function generateInputFields(array $config): string
...
+                $groups = '['.\implode(',', \array_map(function ($value) { return "'$value'"; }, $groups)).']';
```
mcg-web: why not using `var_export` or `json_encode` here?

src/Generator/TypeGenerator.php
murtukov: You means `null !== $groups`

src/Generator/TypeGenerator.php
murtukov: Didn't use `var_export` because it doesn't give a good control over the formatting.

With `var_export` it would output following:
```php
$validator->validate(array (
  0 => 'groups1',
  1 => 'my_group',
));
```
but with my code it outputs:
```php
$validator->validate(['groups1','my_group']);
```
which is more consistent with the rest of the code in the generated classes.

What does `json_encode` have to do with this case?

src/Generator/TypeGenerator.php
mcg-web: Oups, yes that's it

src/Generator/TypeGenerator.php
murtukov: Changed to `json_encode`

Issue #592, opened 2019-10-24
Auto register the resolver maps using service tagging


ste93cry:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | no
| Version/Branch   | 0.13

I was thinking about the reason why the resolver maps are registered manually in the configuration instead of being tagged services that can be injected by the service container directly into the `TypeDecoratorListener` service. The only reason I can think of is that the order in which they are registered can matters, but this can be easily solved using a solution similar to the one of the listeners that can configure their priority. I would be happy to provide a PR that automatically registers the maps. I would also stop supporting the `overblog_graphql.definitions.schema.resolver_maps` option, but this would be a BC so maybe we can just deprecate it. WDYT?

mcg-web:
The bundle supports many schema so resolver map auto registration should also make this configurable (using a service attribute tag?). We can deprecate the option in this version and remove it in the next one. Do you want to work on a PR?

ste93cry:
>Do you want to work on a PR?

Yes, I will try to get something up as soon as possible

PR #615, opened 2019-11-13
Auto-register the resolver maps using service tags

ste93cry:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | maybe, see below
| Deprecations? | yes
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | #592
| License       | MIT

With this PR I'm going to deprecate the `overblog_graphql.definitions.schema.[...].resolver_maps` option and instead we are going to promote the use of the service tag `overblog_graphql.resolver_map`. The tag supports an optional attribute `priority` that allows specifying the priority per-schema of the resolver map

### DIFF
--- a/UPGRADE-0.13.md
+++ b/UPGRADE-0.13.md
@@ -5,6 +5,7 @@ UPGRADE FROM 0.12 to 0.13
 
 - [Rename default_field config](#rename-default_field-config)
 - [Improve default field resolver](#improve-default-field-resolver)
+- [Use service tags to register resolver maps](#use-service-tags-to-register-resolver-maps)
 
 ### Rename default_field config
 
@@ -50,3 +51,27 @@ MyType:
 ```
 
 [see default Field Resolver per type for more details](https://webonyx.github.io/graphql-php/data-fetching/#default-field-resolver-per-type)
+
+### Use service tags to register resolver maps
+
+The resolver maps used to be configured using the `overblog_graphql.definitions.schema.resolver_maps`
+option. This has been deprecated in favour of using service tags to register them.
+
+```diff
+# config/graphql.yaml
+overblog_graphql:
+    definitions:
+        schema:
+             # ...
+-            resolver_maps:
+-                - 'App\GraphQL\MyResolverMap'
+```
+
+```diff
+# services/graphql.yaml
+services:
+-    App\GraphQL\MyResolverMap: ~
++    App\GraphQL\MyResolverMap:
++        tags:
++            - { name: overblog_graphql.resolver_map, schema: default }
+```
--- a/docs/definitions/quick-start.md
+++ b/docs/definitions/quick-start.md
@@ -30,16 +30,6 @@ in files `config/graphql/types/*.graphql`
 
 4. Define schema Resolvers ([more details](resolver-map.md))
 
-```yaml
-# config/packages/graphql.yaml
-overblog_graphql:
-    definitions:
-        schema:
-            # ...
-            resolver_maps:
-                - App\Resolver\MyResolverMap
-```
-
 ```php
 <?php
 
--- a/docs/definitions/resolver-map.md
+++ b/docs/definitions/resolver-map.md
@@ -118,19 +118,21 @@ class MyResolverMap extends ResolverMap
 }
 ```
 
-Declare resolverMap to current schema
+Each resolver map must be tagged with the `overblog_graphql.resolver_map` tag
+that defines at which priority it should run for the given schema. The priority
+is an optional attribute and it has a default value of 0. The higher the number,
+the earlier the resolver map is executed.
 
 ```yaml
-overblog_graphql:
-    definitions:
-        schema:
-            # ...
-            # resolver maps services IDs
-            resolver_maps:
-                - App\Resolver\MyResolverMap
-
+# config/services.yaml
 services:
-    App\Resolver\MyResolverMap: ~
+    App\Resolver\MyResolverMap1:
+        tags:
+            - { name: overblog_graphql.resolver_map, schema: default }
+    
+    App\Resolver\MyResolverMap2:
+        tags:
+            - { name: overblog_graphql.resolver_map, schema: default, priority: 10 }
 ```
 
 **Notes:**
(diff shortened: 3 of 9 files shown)


mcg-web:
Thank you @ste93cry! We'll remove the resolver maps from the configuration in next major version (0.14). I think to ease the multi-schema resolver map configuration we could add a new interface covering this case:

```php

use Overblog\GraphQLBundle\Resolver\ResolverMapInterface;

interface ResolverMapMultiSchemaInterface extends ResolverMapInterface
{
    public static function getSchemaNames(): array;
}
```

this is just an example.

we can also add priority on the main interface so autoconfigure could be use for all cases.

What do you think of this solution?

ste93cry:
Something similar to how the Symfony event subscribers are configured may work, e.g.

```php
public static function getSchemas(): array
{
    // Returns an hashmap where the key is the name of the schema and the value is the priority of the resolver map for that schema
    return [
        'schema_1' => -10,
        'schema_2' => 0,
        'schema_3' => 10,
    ];
}
```

However the tag's configuration is more explicit as it has named arguments. Also, I don't really like to have two interfaces to support two cases that could be unified under the same umbrella, but I understand that editing the `ResolverMapInterface` will break BC and we may want to avoid it (at least for now). In the end, if there are two interfaces it will still not solve the issue with the autoconfiguration because if someone has more than one schema and the resolver maps don't implement the `ResolverMapMultiSchemaInterface` interface then we still don't know which schema applies to each of them unless we assume that they will be added to a certain schema (e.g. the default)

mcg-web:
We can modify the main interface or just check if method exist in the 0.13 and add the method signature in 0.14. BC are acceptable in this case since it is documented.

ste93cry:
> if someone has more than one schema and the resolver maps don't implement the `ResolverMapMultiSchemaInterface` interface then we still don't know which schema applies to each of them unless we assume that they will be added to a certain schema (e.g. the default)

Unless I'm missing something I still don't see how having a method in the interface could help us in not having to specify the schema that the resolver map should be added to when there are more than one. Of course this does not apply if we always require that method to return a filled list so that the user even in the case of one schema has to specify it. Also I fail to see the advantage of one solution (the tag) vs the other (the interface)

mcg-web:
This example:
```php
use Overblog\GraphQLBundle\Resolver;

interface ResolverMapInterface
{
    public static function getSchemas(): array
    //... others methods
}

class MyResolverMap3 implements ResolverMapInterface
{
    public static function getSchemas(): array
    {
        return ['default' => 0, 'private' => 10];
    }
    //... others methods
}
```
is equivalent to

```yaml
services:
    App\Resolver\MyResolverMap3:
        tags:
            - { name: overblog_graphql.resolver_map, schema: default, priority: 0 }
            - { name: overblog_graphql.resolver_map, schema: private, priority: 10 }
```

The only advantage of using interface over config is that it ease configuration and autoconfigure is not require to be disable for resolver map with multiple schema so the service can also be used on existing services that need autoconfigure to configure others tags.

We can merge this one and I can work on this enhancement in a next PR.

mavimo:
If I understand correctly you're discussing about two possible implementations, the "tag based" and "interface based".

## Using tags

### Configuration
#### Mono schema
```yaml
services:
    MyApp\MyResolvers\Resolver1: ~ 
# or
    MyApp\MyResolvers\Resolver2:
        tags:
            { name: overblog_graphql.resolvermap, schema: schema1 }
```
#### Multi schema
```yaml
services:
    MyApp\MyResolvers\Resolver2:
        tags:
            { name: overblog_graphql.resolvermap, schema: schema1 }
        
    MyApp\MyResolvers\Resolver3:
        tags:
            { name: overblog_graphql.resolvermap, schema: schema2 }

    MyApp\MyResolvers\Resolver4:
        tags:
            { name: overblog_graphql.resolvermap, schema: schema1 }
            { name: overblog_graphql.resolvermap, schema: schema2, priority: 10 }
            { name: overblog_graphql.resolvermap, schema: schema3, priority: -3 }
```

## Using interface

### Configuration
#### Mono schema
```yaml
services:
    MyApp\MyResolvers\Resolver1: ~ 
```
#### Multi schema
```yaml
services:
    MyApp\MyResolvers\Resolver2: ~ 
    MyApp\MyResolvers\Resolver3: ~ 
    MyApp\MyResolvers\Resolver4: ~ 
```

Code:

 * Mono schema
```php
namespace MyApp\MyResolvers;

class Resolver1 implements ResolverMapInterface
{

}
```

 * Multi schema
```php
namespace MyApp\MyResolvers;

class Resolver2 implements ResolverMapMultiSchemaInterface
{
    public static function getSchemas(): array
    {
        return [
            'schema1' => 0,
        ];
    }
}

class Resolver3 implements ResolverMapMultiSchemaInterface
{
    public static function getSchemas(): array
    {
        return [
            'schema2' => 0,
        ];
    }
}

class Resolver4 implements ResolverMapMultiSchemaInterface
{
    public static function getSchemas(): array
    {
        return [
            'schema1' => 0,
            'schema2' => -10,
            'schema3' => 3,
        ];
    }
}
```

**JM2C**
I fund the version that use the tag configuration to bind `ResolverMapXX` classes to schema more clear, the only drawback that I see is for the single schema that do not force the user to add the `schema` name, but the version with multiple schema need it, that sound a unclear behaviour (but I'm not sure to know how many projects need to have multiple schema), but this can be solved if this information is well explained in docs.

ste93cry:
As discussed on Slack, I changed the implementation so that the `overblog_graphql.resolver_map` tag is always required and no `autoconfigure` is provided out-of-the-box. However, the build for some reason is failing in the CI but it's passing locally. Any help is appreciated

mcg-web:
This comes from the release of Symfony 5 it add some non-managed deprecation messages. I'll try to fix this today.

ste93cry:
Rebased, tests are passing and what is failing is not related to this PR. Also AppVeyor build is broken, but that's also in `master`

P.S. unrelated note, I'm wondering if it would be possible to somehow speed up the TravisCI build as >20 min per-build is really a lot of time to wait for

mcg-web:
Yes the build is slow but the most important part is done in 5 to 10 minutes, the coverage build is the one taking more time. I merged this PR so I can work on making the master build green again. Thank you @ste93cry !

PR #636, opened 2020-01-13
Only grant permission when `access` returns true

### LINKED ISSUE #635: Validate `access` property
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? |no
| RFC?             |no
| Version/Branch   | v0.12.5

When using the `access` property it's very easy to make mistakes. It wants to have a boolean value or expression language string. But when you forget to prefix your expression with `@=` the string value will result to true, granting access.

Two problems:
1) It should use strict comparison `access === true` and not `access == true`. 
2) It should only accept `true`, `false`, or string starting with `@=`

ruudk:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | could be, if people use `"true"` instead of `true`
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | #635
| License       | MIT

This fixes #635

When you provide an expression inside the `access` option, it will only work as an expression when you prefix it with `@=`. When you forget to do this, it will be just a string.

Previously, the code would check if the access value is trueish instead of strict true. This granted access to all fields that had mistakenly forgotten to prefix with `@=`.

I also flipped the if statement so that it's much easier to parse as a human :D

### DIFF
--- a/src/Resolver/AccessResolver.php
+++ b/src/Resolver/AccessResolver.php
@@ -51,12 +51,12 @@ class AccessResolver
     {
         $promiseOrHasAccess = $this->hasAccess($accessChecker, $resolveArgs);
         $callback = function ($hasAccess) use ($resolveArgs, $resolveCallback) {
-            if (!$hasAccess) {
-                $exceptionClassName = self::isMutationRootField($resolveArgs[3]) ? UserError::class : UserWarning::class;
-                throw new $exceptionClassName('Access denied to this field.');
+            if (true === $hasAccess) {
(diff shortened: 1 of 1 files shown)


mcg-web:
Thank you @ruudk ! Since this is related to security can you submit this PR for 0.11 please? So we can release a security patch for older versions.

ruudk:
@mcg-web Done. Do you have an idea how we can add tests for this?

mcg-web:
Thank you @ruudk !

PR #638, opened 2020-01-20
GraphQL Profiler

Vincz:
| Q             | A
| ------------- | ---
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

This PR add a GraphQL Data Collector allowing to watch the queries and the results.


Vincz:
Hey @mcg-web ! 
Not sure how I could test this [controller](https://github.com/overblog/GraphQLBundle/blob/949bcaaa12e5612913bac73f8c032986f3fc69f7/src/Controller/ProfilerController.php) 
Could use your advice ❤️

mcg-web:
Hey @Vincz, I'm not really sure of how this could be test :thinking: but I'll try to get a solution :smile:

mathroc:
@Vincz I just installed the bundle from your fork and I can see the GraphQL tab on the toolbar and in the profiler, however it says there’s one request on the section title but no requests when trying to show them :man_shrugging: 

image

Vincz:
> @Vincz I just installed the bundle from your fork and I can see the GraphQL tab on the toolbar and in the profiler, however it says there’s one request on the section title but no requests when trying to show them 🤷‍♂
> 
> image

Hey @mathroc :)
To ensure we only get requests on the endpoint, we look in the profiler database for queries with url starting with the endpoint url.

How does resolve the route `overblog_graphql_endpoint` for you ? 
Does it match the one requested when you do your graphql operations?

mathroc:
> Hey @mathroc :)
> To ensure we only get requests on the endpoint, we look in the profiler database for queries with url starting with the endpoint url.
> 
> How does resolve the route `overblog_graphql_endpoint` for you ?
> Does it match the one requested when you do your graphql operations?

screenshot_2020-03-31_11:30:18

that’s what the endpoint resolves to, it looks like it the correct one, however

I took a quick look at the code in the ProfilerController, if I inspect $profile (the `$profile = $this->profiler->loadProfile($token);` one) , I can see the GraphQL query there

but $tokens is empty. maybe it’s because I have a host condition on the endpoint ? anyway, why not using the `$profile` from the current inspected request? how the `$tokens` is built looks like it would show the GraphQL queries of the last 100 symfony requests instead of the currently inspected request :thinking:

Vincz:
@mathroc Yes, I think the problem is related to the parameter in the endpoint url. To build the `$tokens` we use this : 
`$this->profiler->find(null, $this->endpointUrl, '100', null, null, null, null)`
It means that we are querying the profiler requests database for queries where url contains the endpoint path. The reason is that we don't want non GraphQL related queries. It's the only way to filter queries containing GraphQL data that I think off. Maybe, we could allow user to change the "matching" string used to identify GraphQL queries (and set it by default to the endpoint url).

mathroc:
after looking around a bit more, I guess you effectively made a GraphQL queries explorer and inspector that shows GraphQL queries from the last 100 Symfony requests. I can understand why you did that but that’s surprising because it’s not how of other Symfony profilers panels work, they only show collected data about the current request, at least in /_profiler/{token}?panel= . I think it should stay like that at first

the panel you made should be at a url like /_profiler/{token}/graphql (mimicking /_profiler/{token}/search)

---

to collect the Symfony requests where GraphQL have been made, 
instead of ` $this->profiler->find(null, $this->endpointUrl, '100', null, null, null, null)`, you should filter `$this->profile->find(null, null, '100', null, null, null, null)` on `$profile->getCollector("graphql")->getCount() > 0`

but to be clear, I don’t think we should browse across Symfony request (at first at least)

mathroc:
```php
        $tokens = \array_filter(\array_map(function ($tokenData) {
            $profile = $this->profiler->loadProfile($tokenData['token']);
            try {
                $graphql = $profile ? $profile->getCollector('graphql') : null;

                if ($graphql instanceof GraphQLCollector && $graphql->getCount() === 0) {
                    return null;
                }

                $tokenData['graphql'] = $graphql;

                return $tokenData;
            } catch (\InvalidArgumentException $th) {
                return null;
            }
        }, $this->profiler->find(null, null, '100', null, null, null, null)));
```

this makes it work with any endpoint

Vincz:
Your solution works, but it only filter queries after they have been retrieve from the profiler. So, if on the last 100 Symfony requests, only 10 are GraphQL related, you will only display this last 10 GraphQL queries.
The reason it's not a "regular panel" is the usability. In projects using GraphQL, you usually end up with a lot of GraphQL queries (mutation refreshing query, etc...). The default requests panel doesn't show much about the request. So if you are looking for a specific GraphQL operation, you have open each request to find the GraphQL operation you want, it's a painful process. With the current approach, you have all the latest GraphQL operation on one panel, so you can quickly identify the GraphQL request you are looking for. That's the reason I made this feature.

mathroc:
> Your solution works, but it only filter queries after they have been retrieve from the profiler. So, if on the last 100 Symfony requests, only 10 are GraphQL related, you will only display this last 10 GraphQL queries.

ah yes, I didn’t think of that...

> The reason it's not a "regular panel" is the usability. In projects using GraphQL, you usually end up with a lot of GraphQL queries (mutation refreshing query, etc...). The default requests panel doesn't show much about the request. So if you are looking for a specific GraphQL operation, you have open each request to find the GraphQL operation you want, it's a painful process. With the current approach, you have all the latest GraphQL operation on one panel, so you can quickly identify the GraphQL request you are looking for. That's the reason I made this feature.

I understand that very well and that’s definitely something that would be useful, I’m just saying it should appear somewhere else in the profiler (ie: not in a request panel). And for starters a panel for the current request would be awesome already. 

I’m not aware of other profilers having a dedicated page to browse across requests but it should be doable. 
Maybe bits of the profiler search page could be reused for that. Because you’ll like want to have pagination and filters on such a page too.

I’d be happy to help too if there’s interest in going in that direction

Vincz:
We are already using the profiler search, but the problem is that we are very limited as the profiler search in a csv that looks like that:
```
b99b15,127.0.0.1,GET,http://myhost.dev/videos,1581529195,,200
a58951,127.0.0.1,POST,http://myhost.dev/graphql/,1581529196,,200
673d36,127.0.0.1,GET,http://myhost.dev/graphql/,1581529205,b7faf0,500
b7faf0,127.0.0.1,POST,http://myhost.dev/graphql/,1581529205,,500
```
So we can search on ip, method and url, but that's pretty much it. That's why I ended up looking for the endpoint url to be able to extract the relevants queries. Another approach could be to store the GraphQL requests ourselves (in an other csv perhaps?) but it would add a layer of complexity. 

About the dedicated panel, I had one at the beginning but I removed it as it was displaying the exact same info as the ones displayed in the list. When the list works, it highlights the current request in the list (like the router panel displaying the matching route). I also wanted to be quickly able to see the list when you click in the toolbar. Have you been able to have a look at the GraphQL queries list already? What could we add in it? What do you think we could add in a dedicated panel?

mathroc:
> We are already using the profiler search, but the problem is that we are very limited as the profiler search in a csv that looks like that:
> 
> ```
> b99b15,127.0.0.1,GET,http://myhost.dev/videos,1581529195,,200
> a58951,127.0.0.1,POST,http://myhost.dev/graphql/,1581529196,,200
> 673d36,127.0.0.1,GET,http://myhost.dev/graphql/,1581529205,b7faf0,500
> b7faf0,127.0.0.1,POST,http://myhost.dev/graphql/,1581529205,,500
> ```
> 
> So we can search on ip, method and url, but that's pretty much it. That's why I ended up looking for the endpoint url to be able to extract the relevants queries. Another approach could be to store the GraphQL requests ourselves (in an other csv perhaps?) but it would add a layer of complexity.
> 

I’ll try to come up with something with the existing csv and tell you if I’m successful. Storing the GraphQL requests ourselves is probably more trouble than it’s worth imho

> About the dedicated panel, I had one at the beginning but I removed it as it was displaying the exact same info as the ones displayed in the list. When the list works, it highlights the current request in the list (like the router panel displaying the matching route). I also wanted to be quickly able to see the list when you click in the toolbar. Have you been able to have a look at the GraphQL queries list already? What could we add in it? What do you think we could add in a dedicated panel?

yes I’ve seen what it looks like ! :+1: 

If we can find out how much time the request took to resolve that would be nice. Nothing else major comes to mind.

On the UI side, maybe we could try to look more like the others profilers

here is what I came up with by tweaking the Doctrine panel with the browser web inspector:

image

mathroc:
I just noticed that the query name is taken from the first field of the query, eg:

```gql
query me {
  CurrentUser {
  }
}
```

this will show `query CurrentUser`, I was exepecting to have the name `me` shown

either `query me` or `query me.CurrentUser`

also, when insepcting a query from graphiql where I have a lot of queries in the editor, all queries are shown, I expected to see only the executed query

and finally, when there is multiple query in the queryString, it would be nice to be able to see only the gql for this query not the other queries too. I guess that means we’ll lose the initial formatting (and maybe the fragments) but I’d be fine with that

Vincz:
Hey @mathroc :-)
I made an update. You can check if you want. Let me know what you think.

@mcg-web Still no idea about how to test this stuff ?

Screenshot 2020-04-21 at 11 35 31

mathroc:
it looks very nice ! well done :+1: 

I have a small issue, with query such as this one:

```gql
query me {
  Me {
    authenticated
    user {
      email
    }
  }
}


query p {
  Project(id: "ba9698db-44f1-40d5-9a47-ca33b23b01b1") {
    name
  }
}

mutation mut {
  UserUpdateEmail(token: "asdad") {
    id
  }
}
```

if I execute the `me` query, in the profiler the the "small" query looks liks this:

screenshot_2020-04-21_21:53:56

but I executed a query not a mutation and only `Me` was queried, so I think it should show:

screenshot_2020-04-21_21:54:48

I got that by patching the GraphQLCollector:

```php

// in onPostExecutor, I replaced
$batch['graphql'] = $this->extractGraphql($parsed);
// with
$batch['graphql'] = $this->extractGraphql($parsed, $executorArgument->getOperationName());

// and updated extractGraphql like this:
    protected function extractGraphql(DocumentNode $document, ?string $operationName)
    {
        $operation = null;
        $fields = [];

        foreach ($document->definitions as $definition) {
            if ($definition instanceof OperationDefinitionNode) {
                if ($operationName !== $definition->name ? $definition->name->value : "") {
                    continue;
                }

                $operation = $definition->operation;
                foreach ($definition->selectionSet->selections as $selection) {
                    if ($selection instanceof FieldNode) {
                        $name = $selection->name->value;
                        $alias = $selection->alias ? $selection->alias->value : null;

                        $fields[] = [
                            'name' => $name,
                            'alias' => $alias,
                        ];
                    }
                }
            }
        }

        return [
            'operation' => $operation,
            'operationName' => $operationName,
            'fields' => $fields,
        ];
    }
```

---

and I think I found a workaround for the `overblog_graphql_endpoint` search issues, in `ProfilerController`:

```php

        $tokens = [];

        $tokensData = array_merge(
            // finds 100 graphql requests by url
            $this->profiler->find(null, $this->endpointUrl, '100', null, null, null, null),
            // keep the last 100 http request in case they contains GraphQL requests not found earlier
            $this->profiler->find(null, null, '100', null, null, null, null),
            [["token" => $token]] // always search the currently inspected token
        );

        foreach ($tokensData as $tokenData) {
            $profile = $this->profiler->loadProfile($tokenData['token']);
            $graphql = $profile ? $profile->getCollector('graphql') : null;

            if ($graphql instanceof GraphQLCollector && $graphql->getCount() === 0) {
                continue;
            }

            $tokenData['graphql'] = $graphql;

            $tokens[$tokenData['token']] = $tokenData;
        }

```

this way, I can see the GraphQL queries from the 100 last queries (useful because `$this->profiler->find(null, $this->endpointUrl, ...)` does not match if there is a hostname contraint on the route. and it also shows request from the currently inspected token (in case it’s a very old one skipped because of the 100 query limit)

---

last tiny detail, I’m not sure the field names should be underlined in the query summary because they look like links but nothing happens when we click on it. (I tried :laughing: thinking maybe it would expand just this field)

Vincz:
Thank you for your reply @mathroc ! I fixed the multi operations issue, thank you for this ! :-)

About the search issue, I'm not sure about your method, cause it's gonna be the last 100 queries (no matter if they are using GraphQL or not) and to actually check it, we have to load every profiler query. It would be better if we could handle this with a url matching.

Does the custom matching option would work for you? If we could override the endpoint url for the search? 

Something like : 

```
overblog_graphql:
    profiler:
        url_search:   api.coworkees
```

We are very limited to search for tokens in the Profiler storage :-/

mathroc:
ah yes it would be nice with a configuration setting :)

maybe we could add a way to override the limit too (in case it’s too some people would like to increase it)

eg:

```yaml
overblog_graphql:
    profiler:
        search: 
            url: api.coworkees
            limit: 200
```

mathroc:
I have the graphql endpoint and the GraphiQL interface on the same url

* `GET` ⇒ graphiql
* `POST` ⇒ graphql

so maybe the method could be overridden too :angel:

Vincz:
Hey @murtukov ! Can you help me on fixing the tests please ? 
https://ci.appveyor.com/project/overblog/graphqlbundle/builds/33773484
It seems related to the validator features (it also failed on current master). 
It seems you fixed it in the PR you are working on. Can you help me ?

murtukov:
Hi @Vincz 

as I can see in the logs, only tests in PHP Nightly fail and this is ok, because it's PHP 8.0 and the result is ignored.

list_of_tests


If you open PHP 7.3 tests (2042.4 and 2042.5) you can see in the logs at the very bottom a green line, stating that all tests are passed, but it fails anyway, because you are using some deprecated class:

logs


You just need to use `Symfony\Component\String\Inflector\EnglishInflector` instead of `Symfony\Component\Inflector\Inflector` in your `ArgumentsTest::testEvaluator` test. Just ignore PHP Nightly.

P.S. The next version of GraphQLBundle is 1.0 and not 0.14. It will require PHP >= 7.4, which means we don't need tests below 7.4 anymore. Pleaso also read this PR, that I opened today https://github.com/overblog/GraphQLBundle/pull/684

murtukov:
@Vincz I cloned your repository. Will investigate, why it doen't work locally.

Vincz:
Thanks for your quick reply @murtukov !
I can't fix the Inflector because it is used by the Property Accessor component, not directly in our code. 
Also, appveyor fails with PHP 7.4.7, there is some Internal Server Error in the Functional tests 
(see https://ci.appveyor.com/project/overblog/graphqlbundle/builds/33773484).
It seems you fixed it in your own PR (https://ci.appveyor.com/project/overblog/graphqlbundle/builds/33767768).
Thanks again!

murtukov:
@Vincz 

Ah, ok, I remember that issue. I have fixed it in this commit https://github.com/overblog/GraphQLBundle/pull/682/commits/8d36b67c8cdfe7c2ab53cb29a7f4e7a2df5fa495 which is a part of the still opened PR. 

The PR will be opened a wile, so if you don't want to wait, just add this method:
```php
public function __get($name)
{
    return $this->$name ?? null;
}
```
to the bottom of the class `Overblog\GraphQLBundle\Validator\ValidationNode`

Vincz:
@murtukov Awesome ! 👍 
Thank you very much !
I fixed it in my PR.

murtukov:
@Vincz looks like it's still failing. As I said, we don't have to check PHP versions below 7.4 so just wait until we push the new `.travis.yaml` config. I will open a new PR

Vincz:
@murtukov Yes, it's just the deprecated in PHP 7.3. I'll wait for the new travis config ! Thanks again :-)

src/DataCollector/GraphQLCollector.php
```diff
@@ -150,7 +150,7 @@ protected function extractGraphql(DocumentNode $document, ?string $operationName
 
         foreach ($document->definitions as $definition) {
             if ($definition instanceof OperationDefinitionNode) {
-                $definitionOperation = $definition->name ? $definition->name->value : null;
+                $definitionOperation = (null !== $definition->name) ? $definition->name->value : null;
```
murtukov, line 153:
```suggestion
                $definitionOperation = $definition->name->value ?? null;
```

src/Controller/ProfilerController.php
```diff
@@ -0,0 +1,65 @@
...
+use Symfony\Component\Routing\RouterInterface;
+use Twig\Environment;
+
+class ProfilerController
+{
+    private $profiler;
+    private $twig;
+    private $endpointUrl;
+    private $requestExecutor;
+    private $queryMatch;
```
murtukov, line 22:
```suggestion
    private ?Profiler $profiler;
    private ?Environment $twig;
    private string $endpointUrl;
    private RequestExecutor $requestExecutor;
    private ?string $queryMatch;
```

src/Controller/ProfilerController.php
```diff
@@ -0,0 +1,65 @@
...
+
+    public function __construct(Profiler $profiler = null, Environment $twig = null, RouterInterface $router, RequestExecutor $requestExecutor, string $queryMatch = null)
```
murtukov, line 24: Optional parameters cannot stand before required:

```suggestion
    public function __construct(?Profiler $profiler, ?Environment $twig, RouterInterface $router, RequestExecutor $requestExecutor, string $queryMatch = null)
```

src/Controller/ProfilerController.php
```diff
@@ -0,0 +1,65 @@
...
+
+            return $tokenData;
+        }, $this->profiler->find(null, $this->queryMatch ?: $this->endpointUrl, '100', 'POST', null, null, null));
+
+        $schemas = [];
+        foreach ($this->requestExecutor->getSchemasNames() as $schemaName) {
+            $schemas[$schemaName] = SchemaPrinter::doPrint($this->requestExecutor->getSchema($schemaName));
+        }
+
+        return new Response($this->twig->render('@OverblogGraphQL/profiler/graphql.html.twig', [
```
murtukov, line 56: Constructor signature says, that `$twig` can be null, but it's not checked, before calling `$this->twig->render()`

src/DataCollector/GraphQLCollector.php
```diff
@@ -0,0 +1,179 @@
...
+use GraphQL\Error\SyntaxError;
+use GraphQL\Language\AST\DocumentNode;
+use GraphQL\Language\AST\FieldNode;
+use GraphQL\Language\AST\OperationDefinitionNode;
+use GraphQL\Language\Parser;
+use Overblog\GraphQLBundle\Event\ExecutorResultEvent;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\DataCollector\DataCollector;
+
```
murtukov, line 16: I would suggest to remove all all @return tags in the PHPDoc and use type-hints instead. PHP-CS-Fixer will change this automatically in the future.

Vincz:
@murtukov Thanks for the great review. I updated accordingly.

murtukov:
@Vincz required changes are pushed in https://github.com/overblog/GraphQLBundle/pull/685. You have to rebase your branch onto `overblog:master`.

Vincz:
Finally, I implemented some kind of test for the profiler controller. All green now :-)

murtukov:
Hi @Vincz. Although the checks are green, there is a problem: the commits history is mixed with my and @mcg-web's commits, which are already in master. This happened, because instead of rebasing, you merged the `overblog:master` branch into `Vincz:master`. As a result it shows, that this PR changes 186 files and has 75 commits, which is wrong.

What you need to do right now is **interactive rebase**, starting from a commit before all my and @mcg-web's commits. For example you could start interactive rebase from `ba5ebf3` or from `ff4a4ac`. In the process of rebasing you can also squash similar commits, especially related to cs-fixes. After you finish, you have to **force push** the changes to rewrite your commits history. But before you start I would suggest you to make a copy of your working branch, because **rebase** rewrites the git history and could lead to data loss if you mess it up.

In the process of rebasing it probably will show merge conflicts multiple times, which you have to resolve manually.

murtukov:
@Vincz You could also create a new branch, e.g. `Vincz:profiler` from the `overblog:master` and cherry-pick all your commits from `Vincz:master` and apply them into your new branch. Maybe this is even a better option, then rebasing. Then you can open a new PR and close this one in its favor.

Vincz:
@murtukov I'm kind of a Git noob (but learning advanced git usage is on my todo list, I promise 😃 )
What if I create a branch from the current master, copy / paste my files in it, and create a single commit from the diff? Would it be ok?

murtukov:
@Vincz the problem of copy paste is that it will replace the whole files. It would work if you only have created new files in the PR, but you have also changes in already existing files and this probably will create merge conflicts again.

Actually what i suggested (cherry-pick) is similar to copy paste, but it copy-pastes commits that you chose. I don't know what IDE you use, but in PHPStorm there is a function **cherry-pick**, with which you can select commits from the log and paste them in your new branch, thus applying all changes one-by-one. This will preserve all your commits.

**OR**

There is another option - **create a patch** from the chosen commits. In this approach you again need to open git log in your IDE, select all commits relevant to the **Profiler** feature and create a single patch-file from it. Then you can apply this patch in your new branch. 

I'll create this patch file for you, all you need to do is to create a new branch from `overblog:master` and apply this file. If you use PHPStorm you can do it with VCS -> Apply Patch. 

Here is the patch file: https://fil.email/kIKYTMyE

Vincz:
@murtukov Thank you so much ! 
For some reason, the patch wasn't working. It was like if the base source from the patch was different from the current master. So I had errors like that : `can't find file to patch at input line 8`.
Anyway, I recreated a PR from scratch and checked the files one by one.
The new PR is here : https://github.com/overblog/GraphQLBundle/pull/696
I'm closing this one.

Issue #526, opened 2019-07-18
[Annotation] Add option to @Provider to attach to different types


akomm:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | no
| Version/Branch   | ^0.12

Currently you can only provide fields to root types via `@Provider` and `@Mutation`/`@Query` on those root types. AFAIK this is also limited to the [default schema root types](https://github.com/overblog/GraphQLBundle/blob/v0.12.0/docs/annotations/annotations-reference.md#query).

I realized this is also needed for non root types, when you have for example two fields connecting something and this logic requires some different services, I have currently to create a type (service) with two dependencies for different purposes. 

There are maybe different ways of doing it. One could be:
Instead of `@Query` and `@Mutation`, give `@Provider` an option like `type=Query`, which would make the annotated field, etc. being "injected" into the type `Query`. This would also allow it to work with any type and independent of schema (single, multi).

Vincz:
Hi @akomm 
Actually, their is an option `targetType` on `@Query` to target a different type for queries.
See [Query annotation](https://github.com/overblog/GraphQLBundle/blob/master/docs/annotations/annotations-reference.md#query)

akomm:
@Vincz You are right, the I currently have the exact same page open, the description initially confused it as it states it does use the Query from definitions config. :)

Vincz:
The idea of `@Provider` was to be able to tell the parser that this specific class contains queries and/or mutations and be able to "group" them (with a prefix, or with `@Access`). I don't know if I picked the best option to do so. 
But we could add a `queriesTarget` on the `@Provider` to automatically add the `targetType` on queries contained in the Provider. I don't know...

akomm:
It is still a bit different to my proposal, as it is then on per-field base not `@Provider` but might be enough, can not tell yet what is better.

I think the docs should then be something like:
`By default, the resulting field is added to the main Query type (define in configuration at key overblog_graphql.definitions.schema.query)`
instead of:
`The resulting field is added to the main Mutation type (define in configuration at key overblog_graphql.definitions.schema.query).`

akomm:
Oh and missing it on Mutation too. Going to add it there too. :)

Vincz:
That is what I was about to say ^^

akomm:
[...]

PR #639, opened 2020-01-20
Add targetType property to Mutation

snaksa:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | #526 
| License       | MIT

Added targetType property to the Mutation type so that different schemas can be created and mutations can be attached to them.

### DIFF
--- a/src/Config/Parser/AnnotationParser.php
+++ b/src/Config/Parser/AnnotationParser.php
@@ -289,8 +289,22 @@ class AnnotationParser implements PreParserInterface
         $isRootMutation = ($rootMutationType && $gqlName === $rootMutationType);
         $currentValue = ($isRootQuery || $isRootMutation) ? \sprintf("service('%s')", self::formatNamespaceForExpression($reflectionEntity->getName())) : 'value';
 
+        $queryType = $isRootMutation ? 'Mutation' : 'Query';
+
+        if (isset($configs['definitions']['schema'])) {
+            $mutationTypes = [];
+
+            foreach ($configs['definitions']['schema'] as $key => $value) {
+                if (isset($configs['definitions']['schema'][$key]['mutation'])) {
+                    $mutationTypes[] = $configs['definitions']['schema'][$key]['mutation'];
+                }
+            }
+
+            $queryType = \in_array($gqlName, $mutationTypes) ? 'Mutation' : 'Query';
+        }
+
         $gqlConfiguration = self::graphQLTypeConfigFromAnnotation($classAnnotation, $classAnnotations, $properties, $methods, $reflectionEntity->getNamespaceName(), $currentValue);
-        $providerFields = self::getGraphQLFieldsFromProviders($reflectionEntity->getNamespaceName(), $isRootMutation ? 'Mutation' : 'Query', $gqlName, ($isRootQuery || $isRootMutation));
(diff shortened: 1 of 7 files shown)


src/Config/Parser/AnnotationParser.php
```diff
@@ -289,8 +289,22 @@ private static function typeAnnotationToGQLConfiguration(
         $isRootMutation = ($rootMutationType && $gqlName === $rootMutationType);
         $currentValue = ($isRootQuery || $isRootMutation) ? \sprintf("service('%s')", self::formatNamespaceForExpression($reflectionEntity->getName())) : 'value';
 
+        $queryType = $isRootMutation ? 'Mutation' : 'Query';
```
Vincz: I think this part is incomplete and confusing. The Query and Mutation types are considered "root" regardless of the schema, so they should behave the same way (the $currentValue for example). 
Also, if we can target whatever type with the @Query, the @Mutation should only target "root" Mutations. 
I think we should check this and raise an exception if something other than a Mutation is targeted by a `@Mutation`.
Then we should change the isRootQuery & isRootMutation to check in all schema.
And the last part are the default Query / Mutation.
Now, we must decide what we want to target if there is no targetType provided. We used to target the root query and mutation of the `default` schema. Now, we can have multiple schemas and we are not sure that we have one with the name "default". So, my suggestion would be to target the one with name "default" if it exists, and if not the first one of the list of schemas. 
And the last parameter of `getGraphQLFieldsFromProviders` should be rename `$isDefault` as it is used to know if the current type is the default one (the one targeted when no targetType is defined).
The tests should also be updated accordingly.

snaksa: Thank you for your feedback! I will try to implement the changes you mentioned tomorrow. I am new to using this bundle and I have some troubles understanding what do you mean by: **_the @mutation should only target "root" Mutations_**. Do you mean that there might be targetType set to something other than mutation type? And how can a @query target whatever type we choose? This confuses me a little and I guess everything will clear in my head once I understand this concept. Can you guide me to a piece of documentation that explains this and I might have missed?

Vincz: The idea of the @provider is to be able to add query or mutation from various services.
In the case of Query, we can target the "root" Query (ie. The Query define at the key "query" in the schema configuration)...or...any other type (for example, I could target the type "Droid" like in the test fixtures) and the effect will be that I'll have a new queryable field on type Droid. 
Let's say I want to provide a "getVictims" query on the Droid type, it would be queried like that for example:

```
query {
    getDroid(id: 'HKB-3') { // Return a Droid
        getVictims {
            name
        }
    }
}
```

In the case of Mutation, the GraphQL specs says that mutations must be defined on the Mutation object (ie. The "root" mutation define at the key "mutation" in the schema config). So I couldn't target the Droid with my mutation. So the targetType should always be a root Mutation as Mutation are always "root".

Hope it'll help :)

Vincz:
Ping @snaksa ! 
Hey man, any update about this ? :-)

Vincz:
Closing in favor of #710

PR #640, opened 2020-01-21
Added inherits property to type annotation

IceShack:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | - 
| License       | MIT

I added the inherits property to the type annotation as an alternative to the interfaces logic

### DIFF
--- a/docs/annotations/annotations-reference.md
+++ b/docs/annotations/annotations-reference.md
@@ -465,6 +465,7 @@ Optional attributes:
 
 -   **name** : The GraphQL name of the type (default to the class name without namespace)
 -   **interfaces** : An array of GraphQL interface this type inherits from
+-   **inherits** : An array of GraphQL types this type inherits from
 -   **isRelay** : Set to true to have a Relay compatible type (ie. A `clientMutationId` will be added).
 -   **builders**: An array of `@FieldsBuilder` annotations
 
--- a/src/Annotation/Type.php
+++ b/src/Annotation/Type.php
@@ -29,9 +29,9 @@ class Type implements Annotation
     /**
      * Inherited Types.
      *
-     * @var string[]
+     * @var string[]|null
      */
-    public $inherits;
+    public $inherits = null;
 
     /**
      * Is the type a relay payload.
--- a/src/Annotation/Type.php
+++ b/src/Annotation/Type.php
@@ -27,6 +27,13 @@ class Type implements Annotation
     public $interfaces;
 
     /**
+     * Inherited Types.
+     *
+     * @var string[]
+     */
+    public $inherits;
+
+    /**
      * Is the type a relay payload.
      *
      * @var bool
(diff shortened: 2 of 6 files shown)


IceShack:
I also have a branch for 0.12 is that is more interesting
https://github.com/IceShack/GraphQLBundle/tree/feature/inheritsTypeAnnotation0.12

mcg-web:
@Vincz can you review this PR please?

Vincz:
@mcg-web Yes boss ! 😄 
@IceShack Hey man ! Thank for this contribution :) However, I can't find anything in the GraphQL spec or the Webonyx library about types inheritance. Could you provide a link or something about this plz ?

IceShack:
I found it in your documentation (first example):
https://github.com/overblog/GraphQLBundle/blob/master/docs/definitions/type-inheritance.md

And you are right its not part of webonyx, but part of your code ([InheritanceProcessor.php)](https://github.com/overblog/GraphQLBundle/blob/master/src/Config/Processor/InheritanceProcessor.php)

Vincz:
@IceShack You are absolutely right 😺 
The reason I never heard about this feature is that I'm using annotations exclusively. 
With annotation, we are tightly coupled with ... PHP classes. PHP classes do support inheritance already.
That's why, when you need inheritance, you just declare your parent class as a normal PHP class with annotation and your sub classes will extend the parent and inherit it's properties. 
Without annotation, we don't have this option, that's why the 'inherit' option was added. 
I'm not sure it would be a good idea to allow it with annotation as it would give two ways to inherit from parents and the regular PHP way seems more natural to me.
What do you think guys ? @IceShack @mcg-web

mcg-web:
Inheritance for yaml and xml was required to keep thing dry, I'm not using annotations so I can't say if this will be useful in this case.

IceShack:
@Vincz Sorry for my delay. 
I exclusively use annotations aswell.
I couldn't get it to work with just `@GQL\Type()` and the whole interface logic seems a bit too complicated for basic inheritance.
My use case for `inherits` is following:
I safe and cache report data for different types of entities which all have the same attributes (sales,cost,impressions,etc.)
So I have a `abstract class` with the `@ORM\MappedSuperclass()` and `@GQL\Type(name="ReportByRange")` annotations.
The Entity specific class extends the abstract class and only adds the relation to the entity and now has this annotation `@GQL\Type(inherits={"ReportByRange"})`.
With the type interface logic you need to write more Code and duplicate code to accomplish the same thing (taken from the unit tests):
Interface: `@GQL\TypeInterface(resolveType="@=resolver('character_type', [value])")`
Classes: `@GQL\Type(interfaces={"Character"}, resolveField="value")`

Vincz:
Hey @IceShack :) I'm not sure I understand correctly. 
Why can't you add the `@GQL\Field` annotations on the common properties in the abstract class, and the `@GQL\Type` on the child classes ? You just want one single GraphQL type for all your child classes ?
I'm not talking about interfaces or inherits, just plain PHP inheritance.

IceShack:
@Vincz Sorry for the big delay. The plain PHP inheritance only works for protected properties not for private properties.

I opened a new PR #675 for that bugfix.

PR #642, opened 2020-01-21
ArgumentTransformer & validator component

Vincz:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| License       | MIT

Move the code so the dependency to the validator component is optional.


src/Transformer/ArgumentsTransformer.php
```diff
@@ -166,13 +167,13 @@ public function getInstanceAndValidate(string $argType, $data, ResolveInfo $info
                     }
                 }
             }
+
```
murtukov, line 170: It's not showed here, but the line 170 contains spaces, because of which it fails to pass travis job. No empty lines should contain any spaces.

Vincz: Good catch! Thanks ! Fixed ! 👍

murtukov:
@Vincz 👍🏻

murtukov:
@Vincz Seems like all checks finished, but the last process is somehow frozen. Try to trigger checks again, by either pushing a new commit or amending the last one.

PR #643, opened 2020-01-21
Always pass arguments from expression function

kix:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | none
| License       | MIT

Previously, if I was passing arguments to my resolvers with the expression function, I would get `null` instead of an instance if the query has no arguments (e.g. no parens at all in GQL).

However, this breaks resolvers that require an instance argument. This makes `@arguments` always try to return an instance of an InputObject whenever it is required.


murtukov:
ArgumentsTransformer is @Vincz stuff. Ping 📡

PR #663, opened 2020-03-09
Use the exception key in the context of the logger

ste93cry:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| License       | MIT

I noticed that most of the Monolog default handlers uses the `exception` key instead of `throwable` to retrieve the exception from the context data of a log, thus with this PR I'm gonna fix the `ErrorLoggerListener` listener. Here a list of handlers that make this assumption:

- Sentry handler (both v1 and v2)
- Rollbar handler
- New Relic handler

To not break BC both keys could be used and in the next major version the old one could be dropped. Let me know how you want to proceed, and if you want of course.

### DIFF
--- a/src/EventListener/ErrorLoggerListener.php
+++ b/src/EventListener/ErrorLoggerListener.php
@@ -15,65 +15,55 @@ final class ErrorLoggerListener
 {
     public const DEFAULT_LOGGER_SERVICE = 'logger';
 
-    /** @var LoggerInterface */
+    /**
+     * @var LoggerInterface
+     */
     private $logger;
 
-    public function __construct(
-        LoggerInterface $logger = null
-    ) {
-        $this->logger = null === $logger ? new NullLogger() : $logger;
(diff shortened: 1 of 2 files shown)


mcg-web:
Thank you @ste93cry yes we can drop the old one in master.

ste93cry:
Watch out that you have merged a breaking change in `0.12` as I didn't have the time to revert the change of the array key used in the `$context` argument, nor not change the target branch

mcg-web:
I don't get it, can you give me the line concerned please?

ste93cry:
Here we go: https://github.com/overblog/GraphQLBundle/pull/663/files#diff-37ea010fd3ef2da2afa7d805bcd2a5f0R67

mcg-web:
Yes I'll add the `throwable` key to fix the BC and removed it in master. Thanks!

PR #675, opened 2020-05-04
Add private property annotations from inherited classes

IceShack:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | -
| License       | MIT

Annotations from private properties from inherited classes got ignored, because they didn't get checked.


akomm:
@IceShack Out of curiosity: I can't quite follow, why you want the properties private, if they are publicly accessed later? `GQL/Type` means, the `value` injected into resolver of this type will be an instance of the annotated PHP class and the `GLQ/Field` makes the field something you would be publicly able to access via graphql, but it will be invisible in PHP.

IceShack:
@akomm I use getter/setter methods to access properties. Which looks like this:
```php
use Overblog\GraphQLBundle\Annotation as GQL;

class Product
{

    /**
     * @var string
     *
     * @GQL\Field(type="String!")
     */
    private string $name;

    public function getName(): string
    {
        return $this->name;
    }

    public function setName(string $name): Product
    {
        $this->name = $name;

        return $this;
    }
}
```

mcg-web:
I revert this merge on 0.13 since it seem to be more a feature than a bug fix and 0.13 is feature freeze. I'm I wrong @Vincz @IceShack ?

Vincz:
Hey @mcg-web ! It's not a new feature, it's a fix. Previously, annotations on private properties on a parent class were not found (they were supposed to be). This PR fixed this issue.

mcg-web:
Hey @Vincz thanks for your answer so maybe we should make the change from 0.12 if it is a fix related to annotations ?

Vincz:
@mcg-web Yes, definitely 👍

Issue #668, opened 2020-03-14
New PHP code generator


murtukov:
As we know the GraphQL Bundle relies in it's core on an internal PHP code generator library. When I was implementing the validation feature I ran into a problem of editing the generator code. The library is not meant to be scaled and editing it was a hard task. Currently the code under `src/Generator` and `lib/generator` is pretty messy and it's hard to follow, how it works. I am planning to work on other features, which require even more changes of the generator code, that's why I decided to write a new PHP code generator library first. It should clean the current mess, make introduction of new features (depending on the generator) much easier and, if possible, optimize the performance. 

The library is already finished and now I am on the step of integrating it into the bundle. But before I do it, there are some questions, that need to be adressed first. The generator is the core of the bundle and changing it requires attention to other aspects, for example config preprocessing and shape of the config tree, which the new php generator will rely on.

First of all I would like to draw your attention to the config tree of yaml types. Here are some random type configs:

```yaml
DateTime:
    type: custom-scalar
    config:
        description: "The DateTime type"
        serialize: ["Overblog\\GraphQLGenerator\\Tests\\DateTimeType", "serialize"]
        parseValue: ["Overblog\\GraphQLGenerator\\Tests\\DateTimeType", "parseValue"]
        parseLiteral: ["Overblog\\GraphQLGenerator\\Tests\\DateTimeType", "parseLiteral"]

Character:
    type: interface
    config:
        description: "@='A character' ~ ' in the Star Wars Trilogy'"
        fields:
            id:
                type: "String!"
                description: "The id of the character."
            name:
                type: "String"
                description: "The name of the character."
            friends:
                type: "[Character]"
                description: "The friends of the character."
            appearsIn:
                type: "[Episode]"
                description: "Which movies they appear in."
        # used expression language to defined resolver (tagged services)
        resolveType: "Overblog\\GraphQLGenerator\\Tests\\Resolver::resolveType"

Droid:
    type: object
    config:
        description: "A mechanical creature in the Star Wars universe."
        fields:
            id:
                type: "String!"
                description: "The id of the droid."
                resolve: "@=value['id']"
            name:
                type: "String"
                description: "The name of the droid."
            friends:
                type: "[Character]"
                description: "The friends of the droid, or an empty list if they have none."
                resolve: ["Overblog\\GraphQLGenerator\\Tests\\Resolver" , "getFriends"]
            appearsIn:
                type: "[Episode]"
                description: "Which movies they appear in."
            primaryFunction:
[...]

PR #682, opened 2020-06-03
Feature/generator

murtukov:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | #668 
| License       | MIT

Finished integration of the new php code generator ([PHPCodeGenerator](https://github.com/murtukov/PHPCodeGenerator)). Documentation comes later.

The old php code generator is fully removed, there is no `lib` folder anymore. The new generator will be delivered as composer dependency.

**IMPORTANT NOTE**: there were unit tests inside `lib/generator`. Obviously this tests are now removed, since they were meant to test the library itself. I am not sure, if the bundle's tests cover all cases and if this will cause any negative consequences (@mcg-web please review this matter).

This is another big update, so I opened the PR as soon as 100% of the tests passed to reduce the changes of the PR. I still have many TODO's (which you can find in the code) to optimize and reduce the code of the new generator, but it will NOT have any crucial changes. Please, if you have time, pull my fork and test on your machine: https://github.com/murtukov/GraphQLBundle/tree/feature/generator (branch `feature/generator`)

#### Code style & static analysis:
This PR changes the php version to 7.4, which requires a new version of PHP-CS-Fixer, which in its turn applies new code styles. I temporarily commented this new rules in the `.php_cs.dist` and will uncomment them after this PR is merged and will open a separate PR for new code styles and static analysis changes.

#### Saving generated files
@mcg-web I am not sure about all write mode constants that are used by saving the contents into the files. Could you please look, if I used write modes correctly in the file TypeGenerator. One of the constants is never used, I think it’s MODE_DRY_RUN. I created a todo for you called “TODO (mcg-web)”


akomm:
Is >=7.4 not a bit steep? Don't get me wrong, I don't run age old PHP versions either. But currently most stuff is still 7.3 and it runs until 2022. I would probably get to 7.4 much sooner, but considering its public package, should we drop support so early?

Another, secondary thing I'v noticed:

Alpine has quite up to date packages and even bumping to the bleeding age versions would still give max. PHP 7.3. 

This means, the docker test would probably not pass any more. Or it would require some manual fetching/building of php 7.4 instead of the packages. 

While as long as it is possible to run the test without docker, I would not take it as a reason to block progress, but its still worth a consideration.

murtukov:
@akomm making GraphQLBundle 1.0 with PHP >= 7.4 is not my decision. It's stated in [the requirements section](https://github.com/overblog/GraphQLBundle/blob/master/docs/index.md#versions-requirements).

Either way I can refactor code to PHP 7.3 if necessary.

akomm:
If its applied at 1.0 and not before I think its not a problem. I assume its comming around the specified due november this year and not, like tomorrow. I think then 7.4 should be not a problem.

murtukov:
@mcg-web I have cleaned commits, squashed and rebased. You can start review now 🚀

murtukov:
@mcg-web done

PR #686, opened 2020-06-27
Fix PropertyAccess issue

murtukov:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

New versions of the `symfony/property-access` throw exception on object properties created dynamically. This fix adds the magic method `__get()` to the `ValidationNode` class, so that `PropertyAccess` could access dynamic properties.

**Update:**
Also drop PHP versions below 7.4 in `.travis.yaml`. Both issues prevent each other from passing checks, so they need to be pushed together.

PR #687, opened 2020-06-27
Fix ValidationNode.php

murtukov:
Same fix as #686 but preserving older PHP versions


mcg-web:
Thank you @murtukov !

PR #695, opened 2020-07-06
Common project code upgrade

murtukov:
General upgrade of the whole project's code.

#### Upgrade PHPStan to the latest version
  - Uninstall deprecated phpstan-shim
  - Remove all unmatched ignored errors from phpstan-baseline.neon (about 700-800 lines)
  - Rename phpstan.neon.dist (remove .dist suffix) since all contributors have to stick to a single configuration
  - Fix ~700 new errors

#### Upgrade PHPUnit to the latest version
  - Change deprecated test methods accordingly

- Configure new PHP-CS-Fixer rules:
  - Remove all redundant rules, which are already included in Symfony set.
  - Add new rules (now all classes, functions and constants will be automatically shortened and imported. Global functions don't require leading slash anymore)

I also manually went through all files and changed/fixed/tweaked code, added type-hints wherever possible.

#### Upgrade `webonyx/graphql-php` to the latest version
  - Update tests accordingly

#### Change version ranges for Symfony packages:
Change requirements in `composer.json` for all Symfony packages from **`^4.3 || ^5.0`** to **`4.4 - 5.1`**, which is equal to 
**`>=4.4.0 <5.2`**. Also add 2 additional jobs in `.travis.yml` to execute tests with Symfony 5.0 and Symfony 5.1.

#### Note:
With the new PHPStan version errors can be ignored directly in the code [with comments](https://phpstan.org/user-guide/ignoring-errors#ignoring-in-code-using-phpdocs), e.g.:
```php
$name = $this->container()->get('service'); // @phpstan-ignore-line

/** @phpstan-ignore-next-line */
$name = $this->container()->get('service');
```


murtukov:
@mcg-web  @mavimo 

Travis installs wrong dependencies. In `composer.json` the version of `symfony\phpunit-bridge` is `^5.1`, but Travis installs `v4.3.11` and this breaks the test. Do you know, what causes it?

murtukov:
@mcg-web This PR is ready to me merged. It would be nice if we merge it ASAP, because it affects almost all files in the project.

phpstan.neon
```diff
@@ -4,6 +4,8 @@ includes:
 parameters:
   level: 7
   reportUnmatchedIgnoredErrors: false
+  checkMissingIterableValueType: false
+  checkGenericClassInNonGenericObjectType: false
```
murtukov, line 8: This are new options coming with the new version of PHPStan. They indroduce new checks, which we didn't have in previous versions, so I disabled them.

tests/Config/Processor/BuilderProcessorTest.php
```diff
@@ -34,28 +36,23 @@ public function testApiAbuse($name, $type, $builderClass, $exceptionClass, $exce
...
+            ['foo', BuilderProcessor::BUILDER_FIELD_TYPE, 'Fake\Foo', InvalidArgumentException::class, 'Field builder class "Fake\Foo" not found.'],
+            ['foo', BuilderProcessor::BUILDER_FIELDS_TYPE, 'Fake\Foo', InvalidArgumentException::class, 'Fields builder class "Fake\Foo" not found.'],
+            ['foo', BuilderProcessor::BUILDER_FIELD_TYPE, stdClass::class, InvalidArgumentException::class, 'Field builder class should implement "Overblog\GraphQLBundle\Definition\Builder\MappingInterface", but "stdClass" given.'],
+            ['foo', BuilderProcessor::BUILDER_FIELDS_TYPE, stdClass::class, InvalidArgumentException::class, 'Fields builder class should implement "Overblog\GraphQLBundle\Definition\Builder\MappingInterface", but "stdClass" given.'],
+            ['foo', BuilderProcessor::BUILDER_ARGS_TYPE, stdClass::class, InvalidArgumentException::class, 'Args builder class should implement "Overblog\GraphQLBundle\Definition\Builder\MappingInterface", but "stdClass" given.'],
```
murtukov, line 51: Removed entries with `[]` and `null` as third item, because the target method allows only strings anyway (because of type-hint)

src/Definition/ArgumentInterface.php
```diff
@@ -4,14 +4,14 @@
 
 namespace Overblog\GraphQLBundle\Definition;
 
-interface ArgumentInterface extends \ArrayAccess, \Countable
+use ArrayAccess;
+
+interface ArgumentInterface extends ArrayAccess, \Countable
```
mcg-web, line 9: missing use for `Countable`

src/Executor/ExecutorInterface.php
```diff
@@ -12,7 +12,7 @@ interface ExecutorInterface
 {
     /**
      * @param mixed       $rootValue
-     * @param array|null  $contextValue
+     * @param mixed       $contextValue
```
mcg-web, line 15: `array|null|ArrayObject`?

mcg-web:
Allowing webonyx/graphql-php to the latest version should be done from 0.11, 1.0 version will not be release before November this make this a very long run. Same for symfony 5.1.

phpstan.neon
```diff
@@ -4,6 +4,8 @@ includes:
 parameters:
```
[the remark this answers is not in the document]
murtukov: I thought about that, but why would anybody use custom rules, if they eventually need to push changes and follow all rules that are defined in travis?

murtukov:
> Allowing webonyx/graphql-php to the latest version should be done from 0.11, 1.0 version will not be release before November this make this a very long run. Same for symfony 5.1.

I didn't quiet understand. Do I need to change something?

mcg-web:
@murtukov this mean that  we should first allow this version of webonyx/graphql-php and Symfony 5.1 in 0.11 so this change should not be integrated directly to this branch.

composer.json
```diff
@@ -30,14 +30,14 @@
...
         "psr/log": "^1.0",
-        "symfony/config": "^4.3 || ^5.0",
-        "symfony/dependency-injection": "^4.3 || ^5.0",
-        "symfony/event-dispatcher": "^4.3 || ^5.0",
-        "symfony/expression-language": "^4.3 || ^5.0",
-        "symfony/framework-bundle": "^4.3 || ^5.0",
-        "symfony/options-resolver": "^4.3 || ^5.0",
-        "symfony/property-access": "^4.3 || ^5.0",
-        "webonyx/graphql-php": "^0.13.5"
+        "symfony/config": "4.4 - 5.1",
```
mcg-web, line 33: this syntax limit from 4.4.* to 5.1.* vs old syntax that allowed 4.3 to 5.*. Moving from 4.3 to 4.4 is not an issue but limiting the bundle to 5.1 is not what we want here.

murtukov: 5.1 is the last stable version of Symfony. I restricted versions which pass Travis tests. If we allow 5.* it can break apps of users, when a new version of symfony is released, because it will be automatically installed. It is dangerous to allow untested versions to be installed, isn't it?

murtukov:
@mcg-web 

1. Why is it important to change the versions in previous branches **first**? Does it brake something? What happens if we update versions first in master and then in previous versions?
2. How do we allow Symfony 5.1 in 0.11 if its requires Symfony `>= 3.1, <=4.3`?
And `webonyx/graphql-php` 0.13 requires PHP >=7.1. We can maximum upgrade versions on 0.13:

| Version                                                         | PHP        | Symfony          | Support        |
| --------------------------------------------------------------: | ---------: | ---------------: | -------------: |
| [`1.0`](https://github.com/overblog/GraphQLBundle/tree/master)  | `>= 7.4`   | `>= 4.4`         | DEV            |
| [`0.13`](https://github.com/overblog/GraphQLBundle/tree/0.13)   | `>= 7.2`   | `>= 4.3`         | Active support |
| [`0.12`](https://github.com/overblog/GraphQLBundle/tree/0.12)   | `>= 7.1`   | `>= 3.4, <4.4`   | Active support |
| [`0.11`](https://github.com/overblog/GraphQLBundle/tree/0.11)   | `>= 5.6`   | `>= 3.1, <=4.3`  | Active support |

composer.json
mcg-web: Symfony 5 will no have any BC until 6 this mean that the bundle will support it without any major changes.

mcg-web:
Active support mean that we maintain the project until we stop supports. 0.11 already support Symfony 5.*. This doc is outdated, the true reference is composer.json anyway.

composer.json
murtukov: Should I change it to `4.3 - 5.*`?

mcg-web: `4.4 - 5.*` is ok

murtukov: I checked, this syntax is invalid (`4.4 - 5.*`), so i just used your old syntax with 4.4 instead of 4.3

murtukov:
@mcg-web what about the PHP requirements? `webonyx/graphql-php` 0.13 requires PHP >=7.1 and GraphQLBundle 0.11 requires PHP >=5.6

https://github.com/webonyx/graphql-php/blob/master/UPGRADE.md#upgrade-v012x--v013x

mcg-web:
Composer choice between the version that match your php version. 0.11 doesn't force to use 0.13 if your php version is compatible you can use it. The bundle required php minimum 5.6 so 7.1 also answer to this need.

murtukov:
@mcg-web Ok, I understand now.

Will you update webonyx version in previous branches?

mcg-web:
I started working on this from 0.11, missing 8 tests to make green.

src/Relay/Connection/Paginator.php
```diff
@@ -4,45 +4,47 @@
...
         $this->fetcher = $fetcher;
         $this->promise = $promise;
-        $this->connectionBuilder = $connectionBuilder ?: new ConnectionBuilder();
+        $this->connectionBuilder = $connectionBuilder ?? new ConnectionBuilder();
     }
 
     /**
      * @param int|callable $total
      *
-     * @return ConnectionInterface|object A connection or a promise
```
mcg-web, line 36: could we keep `ConnectionInterface|object` object represent promise.

src/Relay/Connection/Paginator.php
```diff
@@ -53,7 +55,7 @@ public function backward(ArgumentInterface $args, $total, array $callableArgs =
     }
 
     /**
-     * @return ConnectionInterface|object A connection or a promise
```
mcg-web, line 56: see previous

src/Relay/Connection/Paginator.php
```diff
@@ -83,7 +85,7 @@ public function forward(ArgumentInterface $args)
     /**
      * @param int|callable $total
      *
-     * @return ConnectionInterface|object A connection or a promise
```
mcg-web, line 86: see previous

src/Relay/Connection/Paginator.php
```diff
@@ -127,11 +130,11 @@ private function handleEntities($entities, callable $callback)
      */
     private function computeTotalCount($total, array $callableArgs = [])
     {
-        if (null !== $this->totalCount) {
```
mcg-web, line 130: why this change? both are not equivalent

src/Request/BatchParser.php
```diff
@@ -6,14 +6,22 @@
...
+use function sprintf;
+use const JSON_ERROR_NONE;
 
 class BatchParser implements ParserInterface
 {
     use UploadParserTrait;
 
     public const PARAM_ID = 'id';
 
-    private static $queriesDefaultValue = [
```
mcg-web, line 16: we can move this to const

src/Relay/Connection/Paginator.php
murtukov: Tested this method with Xdebug and fixed to `Connection|Promise`

src/Relay/Connection/Paginator.php
murtukov: Tested this method with Xdebug and fixed to Connection|Promise

src/Relay/Connection/Paginator.php
murtukov: Tested this method with Xdebug and fixed to Connection|Promise

src/Relay/Connection/Paginator.php
murtukov: In PHP 7.4 it's the same. The property `totalCount` is declared as `private int $totalCount` and can never be null. 
If you try to test like this `null !== $this->totalCount` when the property is not set you will get the error:

```
Uncaught Error: Typed property Test::$total must not be accessed before initialization
```
so it only can either be int or unset.

src/Request/BatchParser.php
murtukov: Done

murtukov:
@mcg-web done 😎

deeky666:
When can we expect to use the new `webonyx/graphql-php` Version? It has been nearly 7 months ago v14 was released... Thank you!

mcg-web:
@deeky666 can you open an issue please ? So we can have some information on the version you using and give you the best answer to your question.

murtukov:
@mcg-web @deeky666 

There is no need to open an issue. The master branch already uses v14. Just need to wait for the next release, after [this PR](https://github.com/overblog/GraphQLBundle/pull/786) is merged.

deeky666:
:+1: thanks! I'll be watchingthe PR then.

mcg-web:
`0.14` will be release in same time with `1.0` since it is just a transient version. `0.14` will have different php and symfony versions constraints, If these doesn't match your needs please open a dedicated issue.

Issue #588, opened 2019-10-19
non-existent parameter "container.build_id"


videni:
hi,  somehow I got this error, any fix please? I'm using Symfony4 and the master branch of this bundle commited at  `7f4be04`. 
```
Symfony\Component\DependencyInjection\Exception\ParameterNotFoundException: You have requested a non-existent parameter "container.build_id". in /Users/john/www/acme-application/vendor/symfony/dependency-injection/ParameterBag/ParameterBag.php:100
Stack trace:
#0 /Users/john/www/acme-application/vendor/symfony/dependency-injection/ParameterBag/EnvPlaceholderParameterBag.php(67): Symfony\Component\DependencyInjection\ParameterBag\ParameterBag->get('container.build...')
#1 /Users/john/www/acme-application/vendor/symfony/dependency-injection/Container.php(112): Symfony\Component\DependencyInjection\ParameterBag\EnvPlaceholderParameterBag->get('container.build...')
#2 /Users/john/www/acme-application/vendor/symfony/dependency-injection/ContainerBuilder.php(1268): Symfony\Component\DependencyInjection\Container->getParameter('container.build...')
#3 /Users/john/www/acme-application/vendor/symfony/dependency-injection/ContainerBuilder.php(1212): Symfony\Component\DependencyInjection\ContainerBuilder->doResolveServices(Object(Symfony\Component\DependencyInjection\Parameter), Array, false)
#4 /Users/john/www/acme-application/vendor/symfony/dependency-injection/ContainerBuilder.php(1114): Symfony\Component\DependencyInjection\ContainerBuilder->doResolveServices(Array, Array, false)
#5 /Users/john/www/acme-application/vendor/symfony/dependency-injection/ContainerBuilder.php(610): Symfony\Component\DependencyInjection\ContainerBuilder->createService(Object(Symfony\Component\DependencyInjection\Definition), Array, false, 'cache.validator')
#6 /Users/john/www/acme-application/vendor/symfony/dependency-injection/ContainerBuilder.php(1264): Symfony\Component\DependencyInjection\ContainerBuilder->doGet('cache.validator', 1, Array, false)
#7 /Users/john/www/acme-application/vendor/symfony/dependency-injection/ContainerBuilder.php(1212): Symfony\Component\DependencyInjection\ContainerBuilder->doResolveServices(Object(Symfony\Component\DependencyInjection\Reference), Array, false)
[...]

PR #697, opened 2020-07-08
Changes setup of expression language functions to fix issue #588

sgehrig:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | #588 
| License       | MIT

As mentioned in comment https://github.com/overblog/GraphQLBundle/issues/588#issuecomment-610393275 the true problem behind issue #588 is that the setup of the expression language functions is flawed. The expression language functions are not designed to take services as constructor arguments. Instead they must rely on dependencies passed into the function during execution.

This PR fixes these issues with several expression language functions. Instead of passing dependencies to the constructor (from where they were injected into closures), the evaluators no rely on the `'globalVariable'` in the `$arguments` parameter - just as the compiler relies on `$globalVariable` variable in the execution context. `'globalVariable'` just points to an instance of `Overblog\GraphQLBundle\Definition\GlobalVariables` (again just the same as for complied function execution). 

Because all of the expression language functions are auto-wired and because the evaluator is only used directly in `Overblog\GraphQLBundle\Validator\Constraints\ExpressionValidator`, there has only been a slight service configuration change (the service `validator.expression` (instance of `Overblog\GraphQLBundle\Validator\Constraints\ExpressionValidator`) now also requires the `Overblog\GraphQLBundle\Definition\GlobalVariables` service as its second constructor argument.

### DIFF
--- a/src/ExpressionLanguage/ExpressionFunction/DependencyInjection/Parameter.php
+++ b/src/ExpressionLanguage/ExpressionFunction/DependencyInjection/Parameter.php
@@ -5,19 +5,18 @@ declare(strict_types=1);
 namespace Overblog\GraphQLBundle\ExpressionLanguage\ExpressionFunction\DependencyInjection;
 
 use Overblog\GraphQLBundle\ExpressionLanguage\ExpressionFunction;
-use Symfony\Component\DependencyInjection\ParameterBag\ParameterBagInterface;
 
 final class Parameter extends ExpressionFunction
 {
-    public function __construct(ParameterBagInterface $parameterBag, $name = 'parameter')
+    public function __construct($name = 'parameter')
     {
         parent::__construct(
             $name,
-            function (string $value) {
+            static function (string $value) {
                 return "\$globalVariable->get('container')->getParameter($value)";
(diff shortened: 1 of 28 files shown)


murtukov:
LGTM too. I've made couple of changes in the master tho, so that `globalVaribales` is not hardcoded. It's used in many places, sometimes plural (globalVariable**s**) and sometimes singular. This change should be applied to the `master` branch taking into account my changes. I can do it.

murtukov:
Thanks you @sgehrig for fixing this annoying issue 👍🏻

sgehrig:
> Thanks you @sgehrig for fixing this annoying issue 👍🏻

Any chance to do a new release for 0.13.3? Oh, and thanks for merging the PR so quick.

murtukov:
@sgehrig all release related questions are for @mcg-web

mcg-web:
done @sgehrig ! Thanks again for the fix!

PR #700, opened 2020-07-09
Treat `useStrictAccess` option as true by default

murtukov:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no

The `useStrictAccess` is a boolean option on GraphQL types. It is `true` by default. There is no need to render both states (true and false), so this PR changes it: it will be rendered only if set to `false`. If it's not defined, it will be considered as `true`.

Here is an example of an object type:
```php
final class HumanType extends ObjectType implements GeneratedTypeInterface
{
    public const NAME = 'Human';
    
    public function __construct(ConfigProcessor $configProcessor, GlobalVariables $globalVariables = null)
    {
        $configLoader = fn() => [
            'name' => self::NAME,
            'description' => 'A humanoid creature in the Star Wars universe.',
            'fields' => fn() => [
                'id' => [
                    'type' => Type::nonNull(Type::string()),
                    'deprecationReason' => 'A terrible reason',
                    'description' => 'The id of the character.',
                    'useStrictAccess' => true,
                ],
                'name' => [
                    'type' => Type::string(),
                    'description' => 'The name of the character.',
                    'useStrictAccess' => true,
                ],
                'friends' => [
                    'type' => Type::listOf($globalVariables->get('typeResolver')->resolve('Character')),
                    'resolve' => fn () => ...,
                    'description' => 'The friends of the character.',
                    'useStrictAccess' => true,
                ],
                'appearsIn' => [
                    'type' => Type::listOf($globalVariables->get('typeResolver')->resolve('Episode')),
                    'description' => 'Which movies they appear in.',
                    'useStrictAccess' => true,
                ],
            ],
        ];
        $config = $configProcessor->process(LazyConfig::create($configLoader, $globalVariables))->load();
        parent::__construct($config);
    }
}
```
Obviously `'useStrictAccess' => true` is not needed here.

Just a quick reminder: the `useStrictAccess` option is set to `false` only if all the following conditions are met:
- The `access` option is set
- The `access` option is an expression
- The `access` expression  contains the `object` variable.

### DIFF
--- a/src/Definition/ConfigProcessor/AclConfigProcessor.php
+++ b/src/Definition/ConfigProcessor/AclConfigProcessor.php
@@ -38,7 +38,7 @@ final class AclConfigProcessor implements ConfigProcessorInterface
                     $field['resolve'] = function ($value, $args, $context, ResolveInfo $info) use ($field, $accessChecker, $accessResolver, $defaultResolver) {
                         $resolverCallback = self::findFieldResolver($field, $info, $defaultResolver);
 
-                        return $accessResolver->resolve($accessChecker, $resolverCallback, [$value, $args, $context, $info], $field['useStrictAccess']);
+                        return $accessResolver->resolve($accessChecker, $resolverCallback, [$value, $args, $context, $info], $field['useStrictAccess'] ?? true);
                     };
                 }
             }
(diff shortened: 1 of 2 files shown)


mcg-web:
Thank you @murtukov for the PR. You should add It's always true for mutations also.

Vincz:
What does the `useStrictAccess` mean ?

mcg-web:
`useStrictAccess` mean that the access does not ACL using the result of the field resolver. In other words, it will not process the field resolver before checking if ACL is ok. In mutation context you can understand why it must always be set as true, in query it helps improve performance.

Vincz:
Oh ok ! Thanks for the explanation.

PR #704, opened 2020-07-09
Update badges (0.11)

murtukov:
Update the top badges to show information about the appropriate version


murtukov:
@mcg-web Seems like Travis hung again without any reason. Could you accept this and other 2 similar PRs anyway?

PR #709, opened 2020-07-10
Allow cleaner access syntax in expressions

murtukov:
This PR adds a magic `__get` method to the `Argument` class, to allow a cleaner syntax in the expressions.

Instead of this:
```console
@=res('my_resolver', [args['firstName'], args['lastName']])
```
We can write this:
```console
@=res('my_resolver', [args.firstName, args.lastName])
```
which is obviously easier to read and is shorter.

And inside resolvers user can do `$args->firstName`, ``$args->lastName``

### DIFF
--- a/src/Definition/Argument.php
+++ b/src/Definition/Argument.php
@@ -87,4 +87,9 @@ class Argument implements ArgumentInterface
     {
         return count($this->rawArguments);
     }
+
+    public function __get(string $name)
+    {
+        return $this->rawArguments[$name] ?? null;
+    }
 }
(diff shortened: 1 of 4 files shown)


src/Definition/Argument.php
```diff
@@ -87,4 +87,12 @@ public function count(): int
     {
         return count($this->rawArguments);
     }
+
+    /**
+     * @return mixed
+     */
+    public function __get(string $name)
```
mcg-web, line 94: Argument is an implementation of ArgumentInterface. This PR break the possibility to use custom Argument implementing ArgumentInterface.

murtukov: I can add `__get` method to the interface too.

mcg-web:
Can you also add an update note please?

murtukov:
What update note? Where?

mcg-web:
In UPGRADE-1.0.md to describe the change done to public interface.

murtukov:
@mcg-web done

mcg-web:
Thank you @murtukov !

Issue #701, opened 2020-07-09
Annotations & Interfaces / Unions auto-discovery


Vincz:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | no
| Version/Branch   | 1.0.0

It could be great to auto-discover a GraphQL Type Unions & Interfaces with annotations.
For a given PHP class we could loop through parents & interfaces.
If one of this parents or interfaces is a GraphQL Interface or Union itself :
- If it's an interface, the interface is added to the GraphQL type 
- If it's an union, the GraphQL type is added to the union 

On `@GQL\Type`, we could had a `autodiscoverInterfaces` option (true by default) and the auto-discovery would only apply if the `interfaces` option is not already set. 

On `@GQL\Union`, we could had a `autodiscoverTypes` option (true by default) and the auto-discovery would only apply if the `types` option is not already set.

murtukov:
@Vincz again the same question: will it be possible with mixed schema (yaml/annotations)?

Vincz:
> @Vincz again the same question: will it be possible with mixed schema (yaml/annotations)?

Why are you doing this to me ??!!! 😭  😜 
You are right, it doesn't make any sense without annotations...

PR #710, opened 2020-07-11
Annotations refactoring

Vincz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? |no
| Tests pass?   | yes
| Documented?   | not yet
| Fixed tickets | #... 
| License       | MIT

- Annotations Refactoring (with a GraphClass)
- Add missing `isTypeOf` on `@GQL\Type`
- Add `targetType` on `@GQL\Mutation` to handle multiple schema #526
- Autodiscover Interfaces for `@GQL\Type` based on PHP interfaces or parent classes #701
- Autodiscover Union types for `@GQL\Union` based on PHP interfaces or children classes #701

### DIFF
--- a/docs/definitions/type-system/object.md
+++ b/docs/definitions/type-system/object.md
@@ -77,6 +77,8 @@ Droid:
 
 ## With Annotations
 
+Note: With annotations, you can omit the `interfaces` option. If so, the system will try to guess the interfaces automaticaly by getting the GraphQL Interface associated with classes that the Class type extends or implements.  
+
 ```php
 <?php
 
--- a/docs/definitions/type-system/object.md
+++ b/docs/definitions/type-system/object.md
@@ -77,7 +77,7 @@ Droid:
 
 ## With Annotations
 
-Note: With annotations, you can omit the `interfaces` option. If so, the system will try to guess the interfaces automaticaly by getting the GraphQL Interface associated with classes that the Class type extends or implements.  
+Note: With annotations, you can omit the `interfaces` option. If so, the system will try to guess the interfaces automatically by getting the GraphQL Interface associated with classes that the Class type extends or implements.  
 
 ```php
 <?php
--- a/docs/definitions/type-system/union.md
+++ b/docs/definitions/type-system/union.md
@@ -16,7 +16,7 @@ HumanAndDroid:
 
 ## With annotations
 
-Note: With annotations, you can ommit the `types` parameter. If so, the system will try to detect GraphQL Type associated with classes that inherit or implement the Union class.  
+Note: With annotations, you can omit the `types` parameter. If so, the system will try to detect GraphQL Type associated with classes that inherit or implement the Union class.  
 
 ```php
 <?php
--- a/docs/definitions/type-system/union.md
+++ b/docs/definitions/type-system/union.md
@@ -16,6 +16,8 @@ HumanAndDroid:
 
 ## With annotations
 
+Note: With annotations, you can ommit the `types` parameter. If so, the system will try to detect GraphQL Type associated with classes that inherit or implement the Union class.  
+
 ```php
 <?php
 
--- a/src/Config/Parser/Annotation/GraphClass.php
+++ b/src/Config/Parser/Annotation/GraphClass.php
@@ -8,7 +8,6 @@ use Doctrine\Common\Annotations\AnnotationException;
 use Doctrine\Common\Annotations\AnnotationReader;
 use Doctrine\Common\Annotations\AnnotationRegistry;
 use ReflectionClass;
-use ReflectionException;
 use ReflectionMethod;
 use ReflectionProperty;
 use RuntimeException;
@@ -41,57 +40,11 @@ class GraphClass extends ReflectionClass
     }
 
     /**
-     * Get an array of parent class names.
-     *
-     * @return string[]
-     */
-    public function getParents(): array
-    {
-        $parents = [];
-        $class = $this;
-        while ($parent = $class->getParentClass()) {
-            $parents[] = $parent->getName();
-            $class = $parent;
-        }
-
-        return $parents;
-    }
-
-    /**
-     * Get the list of methods name.
-     *
-     * @return string[]
-     */
-    public function getMethodsNames(): array
-    {
-        return array_map(fn (ReflectionMethod $method) => $method->getName(), $this->getMethods());
-    }
-
-    public function getMethodAnnotations(string $name): array
-    {
-        return self::getAnnotationReader()->getMethodAnnotations($this->getMethod($name));
-    }
-
-    public function getPropertyAnnotations(string $name): array
-    {
-        return self::getAnnotationReader()->getPropertyAnnotations($this->getProperty($name));
-    }
-
-    /**
      * @return ReflectionProperty[]
      */
     public function getPropertiesExtended()
     {
         return $this->propertiesExtended;
-    }
-
-    public function getPropertyExtended(string $name): ReflectionProperty
-    {
-        if (!isset($this->propertiesExtended[$name])) {
-            throw new ReflectionException(sprintf('Missing property %s on class or parent class %s', $name, $this->getName()));
-        }
-
-        return $this->propertiesExtended[$name];
     }
 
     /**
--- a/src/Config/Parser/Annotation/GraphClass.php
+++ b/src/Config/Parser/Annotation/GraphClass.php
@@ -119,8 +119,8 @@ class GraphClass extends ReflectionClass
     private static function getAnnotationReader(): AnnotationReader
     {
         if (null === self::$annotationReader) {
-            if (!class_exists('\\Doctrine\\Common\\Annotations\\AnnotationReader') ||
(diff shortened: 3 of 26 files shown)


docs/definitions/type-system/object.md
```diff
@@ -77,6 +77,8 @@ Droid:
 
 ## With Annotations
 
+Note: With annotations, you can omit the `interfaces` option. If so, the system will try to guess the interfaces automaticaly by getting the GraphQL Interface associated with classes that the Class type extends or implements.
```
murtukov, line 80:
```suggestion
Note: With annotations, you can omit the `interfaces` option. If so, the system will try to guess the interfaces automatically by getting the GraphQL Interface associated with classes that the Class type extends or implements.  
```

docs/definitions/type-system/union.md
```diff
@@ -16,6 +16,8 @@ HumanAndDroid:
 
 ## With annotations
 
+Note: With annotations, you can ommit the `types` parameter. If so, the system will try to detect GraphQL Type associated with classes that inherit or implement the Union class.
```
murtukov, line 19:
```suggestion
Note: With annotations, you can omit the `types` parameter. If so, the system will try to detect GraphQL Type associated with classes that inherit or implement the Union class.  
```

tests/Config/Parser/fixtures/annotations/Type/Mandalorian.php
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Overblog\GraphQLBundle\Tests\Config\Parser\fixtures\annotations\Type;
+
+use Overblog\GraphQLBundle\Annotation as GQL;
+use Overblog\GraphQLBundle\Tests\Config\Parser\fixtures\annotations\Union\Killable;
+
+/**
+ * @GQL\Type
+ */
+class Mandalorian extends Character implements Killable
```
murtukov, line 13: I think  Mandalorian is not killable 😉

src/Config/Parser/Annotation/GraphClass.php
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        throw new AnnotationException(sprintf('Unable to retrieve annotations from object of class "%s".', get_class($from)));
+    }
+
+    private static function getAnnotationReader(): AnnotationReader
+    {
+        if (null === self::$annotationReader) {
+            if (!class_exists('\\Doctrine\\Common\\Annotations\\AnnotationReader') ||
+                !class_exists('\\Doctrine\\Common\\Annotations\\AnnotationRegistry')) {
```
murtukov, line 123:
```suggestion
            if (!class_exists(AnnotationReader::class) || !class_exists(AnnotationRegistry::class)) {
```

src/Config/Parser/AnnotationParser.php
```diff
@@ -185,67 +175,57 @@ private static function classAnnotationsToGQLConfiguration(
...
-                    self::$providers[$reflectionEntity->getName()] = ['annotation' => $classAnnotation, 'methods' => $methods, 'annotations' => $classAnnotations];
+                    self::$providers[] = ['metadata' => $graphClass, 'annotation' => $classAnnotation];
                 }
                 break;
         }
 
         if (null !== $gqlType) {
             if (!$gqlName) {
-                $gqlName = $classAnnotation->name ?: $reflectionEntity->getShortName();
+                $gqlName = $classAnnotation->name ?: $graphClass->getShortName();
```
murtukov, line 221: I would specify here, of what class is the `$classAnnotation` object is. Either in the docblock or here with something like
```php
/** @var <CLASSNAME> $classAnnotation */
```

src/Config/Parser/AnnotationParser.php
```diff
@@ -257,57 +237,52 @@ private static function classAnnotationsToGQLConfiguration(
...
-
-            $methods = [];
-            foreach ($reflectionEntity->getMethods() as $method) {
-                $methods[$method->getName()] = ['method' => $method, 'annotations' => $annotationReader->getMethodAnnotations($method)];
-            }
-
-            self::$classAnnotationsCache[$className] = [$reflectionEntity, $classAnnotations, $properties, $methods];
+        if (!isset(self::$graphClassCache[$className])) {
+            self::$graphClassCache[$className] = new GraphClass($className);
         }
```
murtukov, line 244: This can be shortened:
```suggestion
        self::$graphClassCache[$className] ??= new GraphClass($className);
```

src/Config/Parser/AnnotationParser.php
```diff
@@ -452,26 +424,43 @@ private static function enumAnnotationToGQLConfiguration(GQL\Enum $enumAnnotatio
...
     /**
      * Get a GraphQL Union configuration from given union annotation.
      */
-    private static function unionAnnotationToGQLConfiguration(string $className, GQL\Union $unionAnnotation, array $classAnnotations, array $methods): array
+    private static function unionAnnotationToGQLConfiguration(GraphClass $graphClass, GQL\Union $unionAnnotation): array
     {
-        $unionConfiguration = ['types' => $unionAnnotation->types];
-        $unionConfiguration = self::getDescriptionConfiguration($classAnnotations) + $unionConfiguration;
+        $unionConfiguration = [];
+        if ($unionAnnotation->types) {
```
murtukov, line 438: Shouldn't it be here `!empty($unionAnnotation->types)` ?

I would generally suggest to use more specific comparison everywhere. It is more save and predictable

murtukov:
@Vincz LGTM, just only one general thing: try to always specify the types in docblocks (param and return) and inline. If you look at the badges from branch 0.11 to master, you will see, that the Scrutinizer reduces with every new version, which means the code quality drops.

For example, if the method returns `array`, add to docblock something like:
- `array<int, MyClass>` - meaning that all keys are integers and all items are objects
- `array<MyClass|YourClass>` - meaning that array contains objects either of class `MyClass` or `YourClass`
- `array<int, MyClass|YourClass>` - same as above, but more specific about keys.

When you specify multiple types like `MyClass|YourClass`, it can create another errors, saying that some methods don't exist. In this case just add inline type with `/** @var <CLASSNAME> $var */` comment.

Any try to fix the errors reported from Scrutinizer. Right now Scrutinizer says, that the 2 worst classes in the whole project are: `TypeBuilder` and `AnnotationParser`. It says, that these files are too big and are too complex and that's true. I am planning to split the `TypeBuilder` (because I created it) into multiple more simple classes and would suggest you to do the same with `AnnotationParser` if that's possible.

Vincz:
Thanks for the great review @murtukov !
I'm not sure how to handle the `/** @var <CLASSNAME> $var */` you suggested when it can have multiple classes. Do you mind to give me an example for this one ? F
I agree about the file size, I already tried to reduced it a bit with the introduction of the `GraphClass` but it's not enough. I'll try to split it also.

Vincz:
For example, with something like this :

```
if (!$fieldType) {
                if ($isMethod) {
                    if ($reflector->hasReturnType()) {
```

I know that because `$isMethod` is true, my `$reflector` is an instance of `ReflectorMethod`. How to make it understand ?

murtukov:
@Vincz 

Example:

```php
/**
 * @param FirstClass|SecondClass $myObject
 */
public function myMethod(object $myObject) 
{
    if($this->type === 'test') {
        $result = $myObject->doSomething();
        return $result;
    }

    $result = $myObject->doAnotherThing();
    return $result;
}
```
Here we say, that `$myObject` can be either of class `FirstClass` or `SecondClass`. Let's say the `doSomething` exists only in `FirstClass` and the `doAnotherThing` exists only in `SecondClass`.

In this case static analyzer will raise an error, saying, that `doSomething` doesn't exist in `SecondClass`, because it doesn't know in which conditions the method `myMethod` will be called. You can fix this by addind 2 comments:

```php
/**
 * @param FirstClass|SecondClass $myObject
 */
public function myMethod(object $myObject) 
{
    if($this->type === 'test') {
        /** @var FirstClass $result */
        $result = $myObject->doSomething();
        return $result;
    }

    /** @var SecondClass $result */
    $result = $myObject->doAnotherThing();
    return $result;
}
```

With that, static analyzers know for sure, what class is used in what line.

murtukov:
@Vincz But this errors maybe wouldn't happend, if we defined param types and return types on all methods we use in the project. Static analyzer can then calculate all possible calls and determine itself, which types are where used.

Here is a useful link: https://phpstan.org/writing-php-code/phpdoc-types

Vincz:
Ok, thanks for the explanation ! Is there a way to let it know the other way around. Like "It is not a ". 
For example here : 

```
if (null !== $gqlType) {
            if (!$gqlName) {
                $gqlName = $classAnnotation->name ?: $graphClass->getShortName();
            }
```

And need to let him know that it is not a `@GQL\Provider` (the only one without ->name).
Or should I explicitly indicate all the possible types ? /** @var First|Second|Third|... $classAnnotation **/

murtukov:
@Vincz you can create an interface `NamedAnnotationInterface` and implement all your annotations with the `name` property from it. Then you can add `/** @var NamedAnnotationInterface $classAnnotation */`.

Issue #712, opened 2020-07-12
Default schema consideration


Vincz:
| Q                | A
| ---------------- | -----
| Bug report?      | maybe

I realised that I've always assumed that the default schema name is "default". But is it a correct assumption ?
If we have multiple schemas like this:

```
schemas:
   schema1:
       ...
   schema2:
       ...
```

Should we consider that there is no default schema or should we consider that the default schema is the first schema defined, ie `schema1`.

What about in this case:
```
schemas:
   schema1:
       ...
   schema2:
       ...
   default:
        ...
```

ping @mcg-web @murtukov

murtukov:
@Vincz why do you need to determine which schema is default?

Vincz:
I needed it in the annotations to get the default types `@Query` and `@Mutation` should apply when no `targetType` is defined (and I take the one with the name `default`).
It also used in the `Executor` to get the schema being executed if none is specified (it takes the first one defined).
And it's also used in the Data collector where it also consider the name 'default' if no schema is set.

murtukov:
@Vincz I would use the first one, if no schema with `default` is defined explicitly

PR #608, opened 2019-11-04
pass schema name to ExecutorArgumentsEvent

videni:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Fixed tickets | #601
| License       | MIT

PR #713, opened 2020-07-12
Add schema name as Executor argument & fix default schema

Vincz:
- Add schemaName in the ExecutorArgumentsEvent
- Make the GraphQL collector retrieve schema from ExecutorArgument instead of routing
- Consider default schema as the one with name 'default' or the first one defined
- Apply same logic to Annotation Parser
- Update doc

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | #608, #712
| License       | MIT

### DIFF
--- a/src/Config/Parser/AnnotationParser.php
+++ b/src/Config/Parser/AnnotationParser.php
@@ -258,15 +258,15 @@ class AnnotationParser implements PreParserInterface
                 $schemaQuery = $schema['query'] ?? null;
                 $schemaMutation = $schema['mutation'] ?? null;
 
-                if ($schemaQuery && $gqlName === $schemaQuery) {
+                if ($gqlName === $schemaQuery) {
                     $isRoot = true;
-                    if ($defaultSchemaName == $schemaName) {
+                    if ($defaultSchemaName === $schemaName) {
                         $isDefault = true;
                     }
-                } elseif ($schemaMutation && $gqlName === $schemaMutation) {
+                } elseif ($gqlName === $schemaMutation) {
                     $isMutation = true;
                     $isRoot = true;
-                    if ($defaultSchemaName == $schemaName) {
+                    if ($defaultSchemaName === $schemaName) {
                         $isDefault = true;
                     }
                 }
--- a/src/Config/Parser/AnnotationParser.php
+++ b/src/Config/Parser/AnnotationParser.php
@@ -253,19 +253,20 @@ class AnnotationParser implements PreParserInterface
     ): array {
         $isMutation = $isDefault = $isRoot = false;
         if (isset($configs['definitions']['schema'])) {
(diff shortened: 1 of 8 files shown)


src/Config/Parser/AnnotationParser.php
```diff
@@ -253,19 +253,20 @@ private static function typeAnnotationToGQLConfiguration(
...
         if (isset($configs['definitions']['schema'])) {
+            $defaultSchemaName = isset($configs['definitions']['schema']['default']) ? 'default' : array_key_first($configs['definitions']['schema']);
             foreach ($configs['definitions']['schema'] as $schemaName => $schema) {
                 $schemaQuery = $schema['query'] ?? null;
                 $schemaMutation = $schema['mutation'] ?? null;
 
                 if ($schemaQuery && $gqlName === $schemaQuery) {
                     $isRoot = true;
-                    if ('default' == $schemaName) {
+                    if ($defaultSchemaName == $schemaName) {
```
mcg-web, line 263: can we use strict check here?

src/DataCollector/GraphQLCollector.php
```diff
@@ -61,8 +65,10 @@ public function getCount(): int
 
     /**
      * Return the targeted schema.
+     *
+     * @return string|false
      */
-    public function getSchema(): string
+    public function getSchema()
```
mcg-web, line 71: this will always return string

src/Config/Parser/AnnotationParser.php
```diff
@@ -253,19 +253,20 @@ private static function typeAnnotationToGQLConfiguration(
...
                         $isDefault = true;
                     }
                 } elseif ($schemaMutation && $gqlName === $schemaMutation) {
                     $isMutation = true;
                     $isRoot = true;
-                    if ('default' == $schemaName) {
+                    if ($defaultSchemaName == $schemaName) {
```
mcg-web, line 269: can we use strict check here?

src/Config/Parser/AnnotationParser.php
murtukov: And a think the "if" conditions could also be simplified with `$schemaQuery === $gqlName` and `$schemaMutation === $gqlName`

src/DataCollector/GraphQLCollector.php
```diff
@@ -28,15 +28,19 @@ public function collect(Request $request, Response $response, Throwable $excepti
     {
         $error = false;
         $count = 0;
+        $schema = false;
```
mcg-web, line 31: use null intead

src/DataCollector/GraphQLCollector.php
mcg-web, line 71: `: ?string`

PR #304, opened 2018-03-23
Update YamlParser.php

AlexGural:
More comfortable to use constant in some configs. Especially in enum type

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| License       | MIT

Add possibility to use native constant expression for `.yml`

Issue #303, opened 2018-03-23
Parsing PHP Constants


AlexGural:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | no
| Version/Branch   | master

Maybe in 
`Overblog\GraphQLBundle\Config\Parser\YamlParser::parce` 
method set parser flag to PARSE_CONSTANT.

` $typesConfig = self::$yamlParser->parse(file_get_contents($file->getPathname()), Yaml::PARSE_CONSTANT);`

Working with enums etc will be more comfortable if we'll have possibility to use constants

mcg-web:
Hi, This is already possible using expression language `'@=constant("\\App\\Foo::Bar")'` but I'm not against a more native implementation. Do you want to submit a PR?

AlexGural:
Why not. I guess it will be more obvious

mcg-web:
Closing this because of inactivity, feel free to continue discussion if this is still needed.

PR #719, opened 2020-07-28
Add more native support for php constants in yaml files

thedustin:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | #303, #304
| License       | MIT

I would like to add "more native" support for PHP constants in YAML files by enabling the `!php/const` tag. 

I struggled a bit while creating my first GraphQL project today, because I expected this feature to be available. Also this way the constants don‘t need to be fetched during runtime with the PHP function `\constant()`.

This topic was already discussed in #303 but closed because of inactivity.

### DIFF
--- a/docs/definitions/type-system/enum.md
+++ b/docs/definitions/type-system/enum.md
@@ -19,7 +19,8 @@ Episode:
                 #deprecationReason: "Just because"
             EMPIRE:
                 # We can use a PHP constant to avoid a magic number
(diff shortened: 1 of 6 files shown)


bravik:
Hi @mcg-web. May I ask are there any plans for releasing this feature?
Merged in July 2020, but I can't find it in latest release which was in December 2020.

PR #720, opened 2020-08-07
fix(field-resolver): re-add support for isser methods

Kocal:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes/no
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | #... 
| License       | MIT

I'm updating from 0.11 to 0.13 on our big app at work, and we have a lot of errors on fields which depend of isser methods.

Isser methods support has been removed in 0.13, because the Symfony PropertyAccessor has been removed too, due to performance issue for large schema.
 
However, I think it can be nice to re-add a basic support out-of-the-box for issers.

What do you think?
Thanks!

### DIFF
--- a/tests/Resolver/ResolverFieldTest.php
+++ b/tests/Resolver/ResolverFieldTest.php
@@ -38,7 +38,7 @@ class ResolverFieldTest extends TestCase
             ['private_property_with_getter2', $object, Toto::PRIVATE_PROPERTY_WITH_GETTER2_VALUE],
             ['not_object_or_array', 'String', null],
             ['name', $object, $object->name],
-            ['enabled', $object, $object->isEnabled()]
+            ['enabled', $object, $object->isEnabled()],
         ];
     }
 }
--- a/tests/Resolver/ResolverFieldTest.php
+++ b/tests/Resolver/ResolverFieldTest.php
@@ -38,6 +38,7 @@ class ResolverFieldTest extends TestCase
             ['private_property_with_getter2', $object, Toto::PRIVATE_PROPERTY_WITH_GETTER2_VALUE],
             ['not_object_or_array', 'String', null],
             ['name', $object, $object->name],
+            ['enabled', $object, $object->isEnabled()]
         ];
     }
 }
--- a/tests/Resolver/Toto.php
+++ b/tests/Resolver/Toto.php
@@ -47,7 +47,8 @@ class Toto
     /**
      * @return bool
      */
-    public function isEnabled() {
+    public function isEnabled()
+    {
         return $this->enabled;
     }
 }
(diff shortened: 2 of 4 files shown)


ruudk:
Will it work when the field is called `isEnabled` and the method is called `function isEnabled() : bool`? By looking at the code, I expect it will try to fetch it as property but I could be wrong.

tests/Resolver/ResolverFieldTest.php
```diff
@@ -38,6 +38,7 @@ public function resolverProvider()
             ['private_property_with_getter2', $object, Toto::PRIVATE_PROPERTY_WITH_GETTER2_VALUE],
             ['not_object_or_array', 'String', null],
             ['name', $object, $object->name],
+            ['enabled', $object, $object->isEnabled()]
```
mcg-web, line 41:
```suggestion
            ['enabled', $object, $object->isEnabled()],
```

tests/Resolver/Toto.php
```diff
@@ -42,4 +43,11 @@ public function resolve()
     {
         return \func_get_args();
     }
+
+    /**
+     * @return bool
+     */
+    public function isEnabled() {
```
mcg-web, line 50:
```suggestion
    public function isEnabled()
    {
```

mcg-web:
@ruudk yes because we test isser before public property.

ruudk:
@mcg-web Maybe this explain it better: https://3v4l.org/tfb8l

ruudk:
Maybe a fix could be to add another `elseif` like this:
```php
} elseif (null !== $getter = self::guessObjectMethod($objectOrArray, $fieldName, '')) { // prefix is empty so it will try `isValid()`
    $value = $objectOrArray->$getter();
}
```

mcg-web:
Maybe the best fix to this issue is to add an optional [ResolverPropertyAccess](https://github.com/overblog/GraphQLBundle/blob/0.12/src/Resolver/Resolver.php) that required `PropertyAccess`. So your both projects will be sure to work like be before 0.13. I want to keep the default as simple as possible, I don't want to rewrite PropertyAccess component here. Can this satisfy both of you @Kocal and @ruudk?

ruudk:
Hmm, then I'd rather have this PR merged, and I will fix these "special" cases on my side.... feels like a missing feature to me but don't want to further complicate this project with having to support 2 types of property access resolvers.

Kocal:
Both options are ok for me

mcg-web:
Ok, I'll merge this one! anyway you can use config to define your custom default field resolver

```
overblog_graphql:
    definitions:
         default_field_resolver:
```

ruudk:
Thanks for merging. Is it too soon to ask for a tagged release 🙈 😂 ?

mcg-web:
> Maybe a fix could be to add another `elseif` like this:
> 
> ```
> } elseif (null !== $getter = self::guessObjectMethod($objectOrArray, $fieldName, '')) { // prefix is empty so it will try `isValid()`
>     $value = $objectOrArray->$getter();
> }
> ```

@ruudk I'm going also to (re-)introduce before release since it is also present on master.

Issue #708, opened 2020-07-10
Resolver definition improvement


murtukov:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | yes
| Version/Branch   | 1.0

These are the next enhancements I would like to work on next.

### Expression language

Currently this bundle heavily (if not completely) relies on Expression Language which is a powerful, yet pretty exotic feature. I don't believe it should be **the main** configuration tool, because it:
 - requires to learn a new pseudo-language to start 
 - has no support of its syntax
 - introduces too much logic into config files

Now it became so ubiquitous in the bundle, that we even use it inside annotations, which itself is a pseudo-language. Consider this piece of code:
```php
/**
 * @GQL\Field(resolve="resolver('hero_friends', [value, args['page']])")
 */
public $friends;
```
the expression inside the annotation looks just ridiculous.

The thing is, that we can achieve same things with more traditional methods and turn the Expression Language from the "must have" to the "nice to have" feature, whaе it actually was invented for. This part is an attempt to rethink the way we define resolvers in our GraphQL schema (one of the ways).

#### Changing the `resolve` to `resolver`
Example:
```yaml
Query:
    type: object
    config:
        fields:
            user:
                type: "User!"
                description: "Something, that talks"
                resolver: App\GraphQL\UserResolver::getUser
```
Just a FQC and method name, no `@=` prefix, no quotes, no double or quadro slashes, no code-eye-parsing. Pretty clean, right? And this syntax is well knows to IDEs, you can just ctrl+click on the link and teleport right to the resolver method... even if it's on the Moon 🚀🌑

The name `resolve` is a verb, which makes sense with Expression Language, but in the new configuration it just points to a method and calling it `resolver` is more suitable here (@akomm's idea btw.). In addition it will allow us to mark the `resolve` field deprecated.

#### Jumping to the resolver method
Now it comes to the resolver configuration. This could be done with the new `@Resolver` annotation. First question that comes into my mind is "**How do we define the order of the resolver params**?". 

Well, first it will try to guess params. We can guess `ArgumentInterface`, `ResolveInfo`, `InputValidator`, `Models` by the type-hints. The parameters without type-hints will be guessed by names: `$value`, `$object`, `$context`, `$user`. 

If non-standart names are used, they could be mapped with the `@Resolver` annotation:
```php
/**
 * Configuring all variables. Only predefined values are available.
 *
[...]

Issue #694, opened 2020-07-06
Allow entity id scalar generation


Vincz:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | no
| Version/Branch   | 1.0.0

I'm currently working on a pretty big corporate application. So I have many entities with cruds, forms, etc... So a lot of code is "automatic" and it's great. I know feel the same need expressed here : https://github.com/overblog/GraphQLBundle/issues/683

The use case is quite simple. When I create a form for an entity (and I have many), I describe my Input with foreign key id as Int and I have to convert them back to entities in my resolver. 
For example:

```yaml
HeroInput:
    type: input-object
    config:
        fields:
            name:
                type: "String!"
           partnerId:
                type: "Int"
````

And in my resolver, I have to do something like : 
```php
function resolve($input) {
   $hero = new Hero();
   $hero->setName($input['name']);
   $partner = $this->getEntityManager()->getRepository(Hero::class)->find($input['partnerId']);
   $hero->setPartner($partner);
}
```
So, I have to be specific for each different entity (one resolver by entity), preventing me from having a single generic resolve function. 

If we could automatically generate scalars to represent each entity, we could save a lot of time. 
The Input would become:
```yaml
HeroInput:
    type: input-object
    config:
        fields:
            name:
                type: "String!"
           partnerId:
                type: "HeroId"
````
and the resolver :
```php
function resolve($input) {
   $hero = new Hero();
   $hero->setName($input['name']);
   $hero->setPartner($input['partner']);
}
```
and the generated scalar would simply be a id to entity converter like https://github.com/overblog/GraphQLBundle/issues/683#issuecomment-650576014 for example.

The option could be opt in by simply adding a "generateScalar" boolean option on the `@GQL\Type` annotation. I'm not sure how we could implement this out of annotation. 

Let me know what you think.

murtukov:
Hi @Vincz,

I am already working for couple of months on a feature called **Hydrator**. It will basically allow you to assotiate a **model** class to an `input-object` type and when a request comes, the model will be populated with data. It is similar to what you created (called `ArgumentsTransformer` if i am correct), but it should be more comprehensive and flexible. One of the **Hydrator**'s features will be ability to apply converters to any field and this is, I believe, what you need right now. I'll show you an example:

```yaml
Mutation:
    type: object
    config:
        fields:
            createHero:
[...]

PR #728, opened 2020-08-13
Injection of `ResolveInfo` into auto-guessed & arguments transformer

Vincz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| License       | MIT

This PR adds a way to inject the `ResolveInfo` through the `Argument Transformers` and will be able to auto-guessed it based on the class `GraphQL\Type\Definition\ResolveInfo`.

Usage example with auto-guessing:

```php
/**
 * @GQL\Type
 */
class MyType {
    /**
     * @GQL\Field(type="[String]!")
     */
    public function getSomething(int $amount, ResolveInfo $info) {
        ...
    }
}
```

Usage example with `@Arg`

```php
/**
 * @GQL\Type
 */
class RootQuery {
    /**
     * @GQL\Field(
     *   type="[User]",
     *   args={
     *     @GQL\Arg(name="ids", type="[Int]"),
     *     @GQL\Arg(name="info", type="@info")
     *   },
     *   resolve="@=call(service('UserRepository').getUser, arguments({ids: '[Int]', info: '@info'}, arg))"
     * )
     */
    public $getUsers;
}

### DIFF
--- a/docs/annotations/arguments-transformer.md
+++ b/docs/annotations/arguments-transformer.md
@@ -100,10 +100,9 @@ class RootMutation {
 
 So, the resolver (the `createUser` method) will receive an instance of the class `UserRegisterInput` instead of an array of data. 
 
-
 ### Special arguments
 
-The arguments transformer is also able to transform some `specials arguments` (see auto-guessing) :
+The arguments transformer is also able to transform some `special arguments` (see auto-guessing) :
 
 - If the type `@info` is specified, it will be replaced by the current `ResolveInfo`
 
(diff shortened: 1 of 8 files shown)


mcg-web:
hi Vinc with the Resolver definition enhancement #708 this PR could be totally obsolete since the way how resolver works has been completely rewritten to be more like a Symfony Controller. Can we wait until before merging this? So we can make annotations takes benefit of this new feature.

Vincz:
> hi Vinc with the Resolver definition enhancement #708 this PR could be totally obsolete since the way how resolver works has been completely rewritten to be more like a Symfony Controller. Can we wait until before merging this? So we can make annotations takes benefit of this new feature.

Yes, you are right, I forgot about it. No problem, we will merge it later only if it is still relevant. 
I was also about to work on a way to extend the auto-guessing (to manage case like : https://github.com/overblog/GraphQLBundle/issues/694), but it's the same and will probably not be relevant after the resolver refactoring.

PR #729, opened 2020-08-15
Add type-hints in annotation classes

murtukov:
### Add type-hints to annotation classes + common refactoring

@Vincz please check the PR carefully, because I don't know everything, that happens in your code, therefore could make mistakes. I just added type-hints to annotation classes and made general changes in the annotation related code. Ask if you have any questions.

### DIFF
--- a/src/Annotation/Access.php
+++ b/src/Annotation/Access.php
@@ -19,5 +19,5 @@ final class Access implements Annotation
      *
      * @var string
      */
-    public $value;
+    public string $value;
 }
--- a/src/Annotation/Mutation.php
+++ b/src/Annotation/Mutation.php
@@ -17,5 +17,5 @@ final class Mutation extends Field
      *
      * @var array<string>
      */
-    public $targetType;
+    public array $targetType;
 }
--- a/src/Annotation/Query.php
+++ b/src/Annotation/Query.php
@@ -17,5 +17,5 @@ final class Query extends Field
      *
      * @var array<string>
      */
-    public $targetType;
+    public array $targetType;
 }
--- a/src/Annotation/Type.php
+++ b/src/Annotation/Type.php
@@ -17,40 +17,40 @@ class Type implements Annotation
      *
      * @var string
      */
-    public $name;
+    public string $name;
 
     /**
      * Type inherited interfaces.
      *
      * @var string[]
      */
-    public $interfaces;
+    public array $interfaces;
 
     /**
      * Is the type a relay payload.
      *
      * @var bool
      */
-    public $isRelay = false;
+    public bool $isRelay = false;
 
     /**
      * Expression to a target fields resolver.
      *
      * @var string
      */
-    public $resolveField;
+    public string $resolveField;
 
     /**
      * List of fields builder.
      *
      * @var array<\Overblog\GraphQLBundle\Annotation\FieldsBuilder>
      */
-    public $builders = [];
+    public array $builders = [];
 
     /**
      * Expression to resolve type for interfaces.
      *
      * @var string
      */
-    public $isTypeOf;
+    public string $isTypeOf;
 }
--- a/src/Config/Parser/AnnotationParser.php
+++ b/src/Config/Parser/AnnotationParser.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Overblog\GraphQLBundle\Config\Parser;
 
+use Doctrine\Common\Annotations\AnnotationException;
 use Doctrine\ORM\Mapping\Column;
 use Doctrine\ORM\Mapping\JoinColumn;
 use Doctrine\ORM\Mapping\ManyToMany;
@@ -68,6 +69,7 @@ class AnnotationParser implements PreParserInterface
      * {@inheritdoc}
      *
      * @throws InvalidArgumentException
+     * @throws ReflectionException
      */
     public static function preParse(SplFileInfo $file, ContainerBuilder $container, array $configs = []): void
     {
@@ -76,6 +78,7 @@ class AnnotationParser implements PreParserInterface
 
     /**
      * @throws InvalidArgumentException
+     * @throws ReflectionException
      */
     public static function parse(SplFileInfo $file, ContainerBuilder $container, array $configs = []): array
     {
@@ -95,8 +98,7 @@ class AnnotationParser implements PreParserInterface
     /**
      * Process a file.
      *
-     * @throws InvalidArgumentException
-     * @throws ReflectionException
+     * @throws InvalidArgumentException|ReflectionException|AnnotationException
      */
     private static function processFile(SplFileInfo $file, ContainerBuilder $container, array $configs, bool $preProcess): array
     {
@@ -140,7 +142,7 @@ class AnnotationParser implements PreParserInterface
         switch (true) {
             case $classAnnotation instanceof GQL\Type:
                 $gqlType = self::GQL_TYPE;
-                $gqlName = $classAnnotation->name ?: $graphClass->getShortName();
+                $gqlName = $classAnnotation->name ?? $graphClass->getShortName();
                 if (!$preProcess) {
                     $gqlConfiguration = self::typeAnnotationToGQLConfiguration($graphClass, $classAnnotation, $gqlName, $configs);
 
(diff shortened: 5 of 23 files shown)


src/Config/Parser/AnnotationParser.php
```diff
@@ -219,7 +223,7 @@ private static function classAnnotationsToGQLConfiguration(
 
         if (null !== $gqlType) {
             if (!$gqlName) {
-                $gqlName = $classAnnotation->name ?: $graphClass->getShortName();
+                $gqlName = isset($classAnnotation->name) ? $classAnnotation->name : $graphClass->getShortName();
```
Vincz, line 226: $gqlName = $classAnnotation->name ?? $graphClass->getShortName();

src/Annotation/Access.php
```diff
@@ -16,8 +16,6 @@ final class Access implements Annotation
      * Field access.
      *
      * @required
-     *
-     * @var string
      */
-    public $value;
+    public string $value;
```
Vincz, line 20: `@var` should not be removed as there are used by Doctrine Annotation for validation. See https://www.doctrine-project.org/projects/doctrine-annotations/en/1.10/custom.html#attribute-types
The properties can still be type hinted.

Vincz:
Does this mean that you started looking into annotations ❤️  ? Or it's just that you get sick of my loose comparisons 😸  ?

src/Annotation/Access.php
murtukov: Are you sure? Because all tests pass. The documentation could be obsolete

src/Config/Parser/AnnotationParser.php
murtukov: I tried it first, but phpstan throws an error "Access to an undefined property object::$name" which is strange, because the `??` operator is a shortcut for `isset` and  `!== null`. This is maybe a bug in phpstan, so I just used `isset` directly to resolve the error. I don't want to flood the code with `// @phpstan-ignore-line` comments, because suppressing errors is the last thing we should do. Let's just leave it `isset` until the bug is fixed.

murtukov:
@Vincz neither of these. I just don't have enough time for complex tasks, so I just go through code here and there and make simple improvements

src/Annotation/Access.php
murtukov: Type-hinted properties cannot be invalid 🙂

Vincz: So, I checked but if we remove the `@var`, Doctrine doesn't do the check anymore and it doesn't use the type hint.
If we remove it and add the type hint, it will work in most of the case but if it is invalid, we get an error like this:
`Typed property Overblog\GraphQLBundle\Annotation\Access::$value must be float, string used`
instead of an error like this:
`[Type Error] Attribute "value" of @GQL\Access declared on property App\Entity\School::$notes expects
   a(n) string, but got boolean.`

For me the second one is more clear. Also, doctrine is able to validate more complexe type like `@var string[]` and we can't check that with type hint.

I think the tests still pass cause we are not testing invalid annotations.

src/Annotation/Query.php
```diff
@@ -17,5 +17,5 @@ final class Query extends Field
      *
      * @var array<string>
      */
-    public $targetType;
+    public array $targetType;
```
mcg-web, line 20: same here

src/Annotation/Type.php
```diff
@@ -14,43 +14,35 @@ class Type implements Annotation
...
-    public $name;
+    public string $name;
 
     /**
      * Type inherited interfaces.
      *
      * @var string[]
      */
-    public $interfaces;
+    public array $interfaces;
```
mcg-web, line 25: same here

src/Config/Parser/AnnotationParser.php
```diff
@@ -149,13 +151,13 @@ private static function classAnnotationsToGQLConfiguration(
...
-                        if (!($classAnnotation->edge xor $classAnnotation->node)) {
-                            throw new InvalidArgumentException(sprintf('The annotation @Connection on class "%s" is invalid. You must define the "edge" OR the "node" attribute.', $graphClass->getName()));
+                        if (!(isset($classAnnotation->edge) xor isset($classAnnotation->node))) {
+                            throw new InvalidArgumentException(sprintf('The annotation @Connection on class "%s" is invalid. You must define either the "edge" OR the "node" attribute, but not both.', $graphClass->getName()));
                         }
 
-                        $edgeType = $classAnnotation->edge;
+                        $edgeType = $classAnnotation->edge ?? false;
                         if (!$edgeType) {
-                            $edgeType = sprintf('%sEdge', $gqlName);
```
mcg-web, line 158: I prefer `sprintf` above concatenation when their not really a matter of performance. `sprintf` it more readable and can give some more information like `%d` for example say that we waiting for an int value here.

src/Annotation/Mutation.php
```diff
@@ -17,5 +17,5 @@ final class Mutation extends Field
      *
      * @var array<string>
      */
-    public $targetType;
+    public array $targetType;
```
mcg-web, line 20: we should keep the typehint `string[]` it give an useful information.

src/Annotation/Access.php
murtukov: Makes sense. But I didn't remove `@var` annotation for arrays.

src/Annotation/Mutation.php
murtukov: `array<string>` and `string[]` are equal.

src/Annotation/Type.php
murtukov: I didn't understand, what is wrong here?

src/Annotation/Mutation.php
mcg-web: The first diff was removing it not replacing it. But yes I know both are the same.

src/Annotation/Type.php
mcg-web: Wrong diff in the first place, you can forget it.

src/Annotation/Mutation.php
murtukov: I didn't remove it. I never remove annotations, that provide extra information.

src/Config/Parser/AnnotationParser.php
murtukov: @mcg-web @Vincz 

Concatenation (or interpolation) always converts value to string, exactly as the `%s` placeholder does. 
So the `sprintf('%sEdge', $gqlName)` doesn't eventually provide more information, than concatenation.

I know 2 scenarios, when `sprintf`makes sense:

1) When using placeholders different from `%s`, for example `%d`, `%b`, `%f` etc. We rarely use `sprintf` with other placeholder types. Currently we have in this project 215 `sprintf`'s with `%s` and only 2 with `%d`. And this is after I replaced many `sprintf` calls with concatenation. 

2) When variables are too long or require extra inline calculation, e.g.:
```php
sprintf("service('%s')", self::formatNamespaceForExpression($graphClass->getName()))
```
---

In all other cases it just brings an unnecessary overhead (function call), because I dont see, how this:
```php
sprintf('Hello, my name is %s and i am %s years old. I am from %s.', $name, $age, $place);
```
is more readable, than this:
```php
"Hello, my name is $name and i am $age years old. I am from $place.";
```
or this:
```php
"Hello, my name is {$name} and i am {$age} years old. I am from {$place}.";
```

I am against using of language constructs, unless it's really necessary or it add's some readability. I think we overuse the `sprintf` function. Look at the line 470 (file `AnnotationParser.php`):
```php
throw new InvalidArgumentException(sprintf('The "resolveType()" method on class must be static and public. Or you must define a "resolveType" attribute on the @Union annotation.'));
```
Why do we have here `sprintf`? It doesn't even have placeholders.

mcg-web: For `AnnotationParser.php` it should not be used here, its a mistake, we can safely removed it. It not about what you  against here, we just try to follow [Symfony Coding standard](https://symfony.com/doc/current/contributing/code/standards.html) 
> Exception and error message strings must be concatenated using sprintf;

When doing opensource we should always try to have and follow some standards, the only accepted reasons of bypassing this is when its mater of performance or security.
Changing standards can always be done but it should be in separates PRs to ease reviewer works and stay focus on PR main issue.

src/Annotation/Access.php
Vincz: So are we sure we want to type hint the annotations properties ? 
If we make a type mistake in an annotation, we will get a less explicit errors as the `Typed property` error will be raised instead of the usual annotation error.

This kind of message `Typed property Overblog\GraphQLBundle\Annotation\Access::$value must be float, string used` doesn't really help to locate the problem.

Vincz: And usually, end users do not manipulate this annotation classes.

murtukov: > If we make a type mistake in an annotation, we will get a less explicit errors as the Typed property error will be raised instead of the usual annotation error.

Not true. Before data is actually stored in the annotation class it's first parsed and validated by AnnotationReader, therefore validation will be performed first.

Vincz: Oh so sorry then. I tested it quick but I guess I had removed the `@var` before testing.

src/Config/Parser/AnnotationParser.php
murtukov: Symfony Coding Standards says only 1 sentence about `sprintf`:
> Exception and error message strings must be concatenated using sprintf;

This example is neither exception nor error message:
```php
$edgeType = sprintf('%sEdge', $gqlName);
```

murtukov: @mcg-web 

> It not about what you against here

Well, you said "I prefer `sprintf`" and a said "I am against `sprintf`" 😀
There are no standarts about `sprintf`, except when using for exceptions and errors.

mcg-web: "I prefer" was just a way of being polite and opening the way to discussion but next time I can be more straightforward if this can help to go faster :+1:

murtukov: I prefer straightforwardness 🤖

mcg-web: :smile: :+1:

Issue #674, opened 2020-04-20
Multiple field and access


Vincz:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | yes
| Version/Branch   | 0.13.2

When using annotations, if we found a multiple relation with doctrine (OneToMany or ManyToMany), we generate a non null multiple (like: `[Friends]!`). It means that the field will always be an array (but it can be empty). So we have a `Type::nonNull` field. 
If we have an `access` on this field, like `@=isGranted()` and we query for this field without being allowed, it will return a null value for the entire object instead of just for this fields itself (I haven't been able to really trace the problem, but I think it is in webonyx/graphql). 

There is multiple ways to solve this problem. 
1) Generate nullable type instead of non nullable type
2) Detect if we have an "access" on a field and if the field is "multiple", remove the non nullable
3) Try to detect the case in the request handling and return an empty array instead of a null value

Personally, I would prefer the second one as if we have a field with an `access`, we are supposed to be able to return `null` for this field.

What do you think about this guys ? 
ping @mcg-web @murtukov @akomm  ?

I make a PR.

murtukov:
Hi, @Vincz. 

Why are you trying to fix the side effect cause by another bug? If I understand correctly, the `access` option should not return null for the entire object and this is the bug we should fix. The 3 options that you suggest look like a temporary hack.

Vincz:
In fact, I'm not even sure it is a bug. 
I checked the code of webonyx on fields resolution, and when a field is marked as not null, if the resolution failed (ie. An exception is throw) for this field, the entire operation failed and we get a `null` for the entire operation. I think that's because we are not supposed to have exception raised during resolution. See this [line](https://github.com/webonyx/graphql-php/blob/6829ae58f4c59121df1f86915fb9917a2ec595e8/src/Executor/ReferenceExecutor.php#L660). Note that it seems to have change in the non published version of webonyx/graphql

The way the `access` stuff works in the bundle, is by adding a resolver that raise an exception if the access is denied, and it's in the webonyx lib that the exception is handled.

Maybe we should change the logic on our side, I don't know.
By the way, what is exactly the "concept" behind our access feature. 
If I query a field I don't have access to, should I get an error or should I get an empty value for this field. Or should it be an option?

mcg-web:
[...]

PR #731, opened 2020-08-15
Access denied logic

Vincz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | #674 
| License       | MIT

This PR adds the ability to return `null` instead of raising an exception when the access to a field is denied.
It will not enforce the field as `nullable` so field with an `access` must be configured properly. 

@mcg-web I didn't force the field as `nullable` as it implied to modify the user defined schema and would for example modify all a type fields as nullable if the object has a `fieldsDefaultAccess`. Better to let the user implement the logic himself. Don't you think?


docs/security/fields-access-control.md
```diff
@@ -6,6 +6,12 @@ An access control can be added on each field using `config.fields.*.access` or g
 If `config.fields.*.access` value is true field will be normally resolved but will be `null` otherwise.
 Act like access is`true` if not set.
 
+Additional configuration can be set using `config.fields.*.accessConfig` or globally with `config.fieldsDefaultAccessConfig`.  
+The available option is :
+
+`nullOnDenied` : Boolean indicating if we should return `null` instead of raising an exception is the access to the field is denied. `default=false`.  **Warning: When using this option make sure the corresponding field is nullable.**
```
murtukov, line 12:
```suggestion
`nullOnDenied` : Boolean indicating if we should return `null` instead of raising an exception if the access to the field is denied. `default=false`.  **Warning: When using this option make sure the corresponding field is nullable.**   
```

docs/security/object-access-control.md
```diff
@@ -8,7 +8,7 @@ An access control can be added on the whole object using a decorator type for th
 
 An access control can be added on each field using `config.fields.*.access` or globally with `config.fieldsDefaultAccess`.
 If `config.fields.*.access` value is true field will be normally resolved but will be `null` otherwise.
-Act like access is`true` if not set.
+Act like access is `true` if not set.
```
murtukov, line 11:
```suggestion
Acts like access is `true` if not set.
```

src/Annotation/Access.php
```diff
@@ -15,9 +15,14 @@ final class Access implements Annotation
...
      * @var string
      */
     public $value;
+
+    /**
+     * Should we return null in access denied.
+     *
+     * @var bool
+     */
+    public $nullOnDenied;
```
murtukov, line 27:
```suggestion
    public bool $nullOnDenied;
```

src/Config/ObjectTypeDefinition.php
```diff
@@ -67,6 +73,29 @@ private function treatFieldsDefaultAccess(ArrayNodeDefinition $node): void
...
+     */
+    private function treatFieldsDefaultAccessConfig(ArrayNodeDefinition $node): void
+    {
+        $node->validate()
+            ->ifTrue(function ($v) {
+                return array_key_exists('fieldsDefaultAccessConfig', $v) && null !== $v['fieldsDefaultAccessConfig'];
+            })
+            ->then(function ($v) {
+                foreach ($v['fields'] as &$field) {
+                    if (array_key_exists('accessConfig', $field) && null !== $field['accessConfig']) {
```
murtukov, line 87:
```suggestion
                    if (isset($field['accessConfig'])) {
```
The `isset` operator on arrays checks, if the key exists and if it's not null

src/Resolver/AccessResolver.php
```diff
@@ -29,24 +28,27 @@ public function __construct(PromiseAdapter $promiseAdapter)
...
+        $resultOrPromise = $resolveCallback(...$resolveArgs);
 
         if ($this->isThenable($resultOrPromise)) {
             return $this->createPromise(
                 $resultOrPromise,
-                function ($result) use ($accessChecker, $resolveArgs) {
-                    return $this->processFilter($result, $accessChecker, $resolveArgs);
+                function ($result) use ($accessChecker, $resolveArgs, $nullOnDenied) {
+                    return $this->processFilter($result, $accessChecker, $resolveArgs, $nullOnDenied);
                 }
```
murtukov, line 47:
```suggestion
                fn ($result) => $this->processFilter($result, $accessChecker, $resolveArgs, $nullOnDenied)
```

src/Resolver/AccessResolver.php
```diff
@@ -57,14 +59,16 @@ private static function isMutationRootField(ResolveInfo $info): bool
     /**
      * @return Promise|mixed
      */
-    private function checkAccessForStrictMode(callable $accessChecker, callable $resolveCallback, array $resolveArgs = [])
+    private function checkAccessForStrictMode(callable $accessChecker, callable $resolveCallback, array $resolveArgs = [], bool $nullOnDenied = false)
     {
         $promiseOrHasAccess = $this->hasAccess($accessChecker, $resolveArgs);
-        $callback = function ($hasAccess) use ($resolveArgs, $resolveCallback) {
+        $callback = function ($hasAccess) use ($resolveArgs, $resolveCallback, $nullOnDenied) {
             if (true === $hasAccess) {
```
murtukov, line 66:
```suggestion
            if ($hasAccess) {
```
No need to check boolean variables with `===` in `if` conditions

tests/Config/Parser/fixtures/annotations/Repository/WeaponRepository.php
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * @GQL\Provider
+ * @GQL\Access("default_access", nullOnDenied=true)
+ */
+class WeaponRepository
+{
+    /**
+     * @GQL\Query
+     */
+    public function searchSecretWeapon(): ?bool
```
murtukov, line 18:
```suggestion
    public function searchSecretWeapon(): bool
```

tests/Functional/App/config/access/mapping/access.types.yml
```diff
@@ -8,15 +8,15 @@ RootQuery:
                 resolve: '@=resolver("query")'
             youShallNotSeeThisUnauthenticated:
                 type: SecureField
-                access: '@=isFullyAuthenticated()'
-                resolve: '@=[]'
+                access: "@=isFullyAuthenticated()"
```
murtukov, line 11: Is there a reason why you replaced single quotes with double quotes?

tests/Functional/Security/AccessTest.php
```diff
@@ -238,6 +238,67 @@ public function testUserForbiddenField(): void
...
+            ],
+        ];
+
+        $query = <<<'EOF'
+query MyQuery {
+  user {
+    nullField
+  }
+}
+EOF;
```
murtukov, line 257:
```suggestion
        $query = <<<GQL
        query MyQuery {
          user {
            nullField
          }
        }
        GQL;
```
There is no need to align EOF strings to the left side, cause PHP automatically removes unnecessary spaces according to the closing EOF. It also makes sense to use some meaningful text, describing the string, instead of just generic `EOF`. Also the quotes in the `'EOF'` can be removed, if there is no variable interpolation.

tests/Functional/Security/AccessTest.php
```diff
@@ -238,6 +238,67 @@ public function testUserForbiddenField(): void
...
+            ],
+        ];
+
+        $query = <<<'EOF'
+query MyQuery {
+  user {
+    promiseNullField
+  }
+}
+EOF;
```
murtukov, line 278:
```suggestion
        $query = <<<GQL
        query MyQuery {
          user {
            promiseNullField
          }
        }
        GQL;
```
same here

tests/Functional/Security/AccessTest.php
```diff
@@ -238,6 +238,67 @@ public function testUserForbiddenField(): void
...
+            ],
+        ];
+
+        $query = <<<'EOF'
+query MyQuery {
+  user {
+    promiseNullFieldNotStrict
+  }
+}
+EOF;
```
murtukov, line 298:
```suggestion
        $query = <<<GQL
        query MyQuery {
          user {
            promiseNullFieldNotStrict
          }
        }
        GQL;
```
same

docs/security/fields-access-control.md
```diff
@@ -6,6 +6,12 @@ An access control can be added on each field using `config.fields.*.access` or g
 If `config.fields.*.access` value is true field will be normally resolved but will be `null` otherwise.
 Act like access is`true` if not set.
```
murtukov, line 7:
```suggestion
Acts like access is`true` if not set.
```

murtukov, line 7:
```suggestion
If `config.fields.*.access` value is true, field will be resolved normally, otherwise will be `null`.
Acts like `access` is`true` if not set.
```

src/Config/ObjectTypeDefinition.php
```diff
@@ -58,7 +65,6 @@ private function treatFieldsDefaultAccess(ArrayNodeDefinition $node): void
                     if (array_key_exists('access', $field) && null !== $field['access']) {
```
murtukov, line 65:
```suggestion
                    if (isset($field['access'])) {
```
The isset operator checks, if the key exists and is not null

src/Config/ObjectTypeDefinition.php
```diff
@@ -67,6 +73,29 @@ private function treatFieldsDefaultAccess(ArrayNodeDefinition $node): void
...
     }
 
+    /**
+     * set empty fields.accessConfig with fieldsDefaultAccessConfig values if is set?
+     */
+    private function treatFieldsDefaultAccessConfig(ArrayNodeDefinition $node): void
+    {
+        $node->validate()
+            ->ifTrue(function ($v) {
+                return array_key_exists('fieldsDefaultAccessConfig', $v) && null !== $v['fieldsDefaultAccessConfig'];
```
murtukov, line 83:
```suggestion
                return isset($v['fieldsDefaultAccessConfig']);
```
The isset operator checks, if the key exists and is not null

src/Resolver/AccessResolver.php
```diff
@@ -29,24 +28,27 @@ public function __construct(PromiseAdapter $promiseAdapter)
     /**
      * @return Promise|mixed|Connection
      */
-    public function resolve(callable $accessChecker, callable $resolveCallback, array $resolveArgs = [], bool $useStrictAccess = false)
+    public function resolve(callable $accessChecker, callable $resolveCallback, array $resolveArgs = [], array $accessConfig = [])
     {
+        $useStrictAccess = $accessConfig['useStrictAccess'];
```
mcg-web, line 33: please inject variables, array make it more difficult to debug since we loose exact  variables list or type here

src/Resolver/AccessResolver.php
mcg-web: `true === $hasAccess` is a matter of security we don't want wrong return types (string, object, ...) to give access to a protected field.

tests/Functional/Security/AccessTest.php
mcg-web: `<<<'GQL'` we don't have vars here

tests/Functional/Security/AccessTest.php
mcg-web: <<<'GQL' we don't have vars here

tests/Functional/Security/AccessTest.php
mcg-web: <<<'GQL' we don't have vars here

src/Annotation/Access.php
Vincz: Still not sure about annotation properties type hint as suggested on #729

src/Resolver/AccessResolver.php
Vincz: Done

src/Resolver/AccessResolver.php
murtukov: I thought $hasAccess is `bool`. I would change the naming of this variable, because it's confusing. Or we can allow only boolean values for `hasAccess`

Vincz: Yes, we could enforce it as a boolean I think. 
It's just that if people used to return non boolean value from access callback, their code will break (but if we target v1.0.0, it's ok right ?).
So `hasAccess` should return a boolean or a promise resolving to a boolean. 
Maybe we could check the return type of the callback and the promise and raise an explicit exception if it's not a boolean ?

tests/Functional/Security/AccessTest.php
```diff
@@ -238,6 +238,67 @@ public function testUserForbiddenField(): void
...
+            ],
+        ];
+
+        $query = <<<'GQL'
+query MyQuery {
+  user {
+    nullField
+  }
+}
+GQL;
```
murtukov, line 257:
```suggestion
        $query = <<<'GQL'
        query MyQuery {
          user {
            nullField
          }
        }
        GQL;
```

tests/Functional/Security/AccessTest.php
```diff
@@ -238,6 +238,67 @@ public function testUserForbiddenField(): void
...
+            ],
+        ];
+
+        $query = <<<'GQL'
+query MyQuery {
+  user {
+    promiseNullField
+  }
+}
+GQL;
```
murtukov, line 278:
```suggestion
        $query = <<<'GQL'
        query MyQuery {
          user {
            promiseNullField
          }
        }
        GQL;
```

tests/Functional/Security/AccessTest.php
```diff
@@ -238,6 +238,67 @@ public function testUserForbiddenField(): void
...
+            ],
+        ];
+
+        $query = <<<'GQL'
+query MyQuery {
+  user {
+    promiseNullFieldNotStrict
+  }
+}
+GQL;
```
murtukov, line 298:
```suggestion
        $query = <<<'GQL'
        query MyQuery {
          user {
            promiseNullFieldNotStrict
          }
        }
        GQL;
```

src/Resolver/AccessResolver.php
murtukov: @Vincz I meant the argument `$hasAccess` should be bolean, so that we don't do: `true === $hasAccess`. But anyway, this is not a part of this PR so I just mark this conversation as resolved

Vincz:
@mcg-web Should we raise an exception when non boolean valus are returned from access callback or if the promise resolve to non boolean value ? Right now, if something other than `true` is return, it's an access denied and we raise an exception or a warning.

murtukov:
@Vincz squashing commits after all tests pass would be nice

Vincz:
@murtukov I'll try ! But I'm waiting for @mcg-web to complete is review as it's security related.

mcg-web:
After re-reading this PR I don't really get it. 
Lets take this example:
```yaml
Query:
  type: object
  config:
    fields:
        user: User!

User:
  type: object
  config:
    fields:
        nullableField:
          type: String
          access: #...
       nonNullableField:
          type: String!
          access: #...
```

if an user without access request `nullableField`
```graphql
{ user { nullableField }
```
the result will be null with an access warning
```json
{ "data": { "user": { "nullableField": null } }, "extensions": {"warnings": ["..."]} }
```

but if this same user request `nullableField`
```graphql
{ user { nonNullableField }
```

the result will be an error since the field can't be nullable
```json
{ "data": null, "errors": ["..."] }
```

 @Vincz What the difference with your PR ? Can you provide an example please.

Vincz:
@mcg-web You are right. The only difference is that we don't get the warning when the field is nullable. It doesn't make sense without the nullable part. I'm sorry. I don't think clearly this day due to too much work. I need vacation. I'll have a break and check this later if I can address the nullable problem. Sorry again.

mcg-web:
All this part of the bundle is not really clear, that's why in the next version we must try to make this easier. Not a big deal sorry for the time I take before seeing that this was not totally complete :disappointed_relieved: .

PR #734, opened 2020-08-16
Revert "Allow to use non-empty constructor for input types"

mcg-web:
Reverts overblog/GraphQLBundle#656


Vincz:
@mcg-web So the idea is that the `Arguments Transformer` will try to populate Input classes with incoming data. Usually, this Input classes just serve as data objects to handle validation. So, the `Arguments Transformer` used to instantiate with `new InputClass()` and It breaks if there is required parameters in the constructor. Instantiating the input class without the constructor prevents this.
But I agree, it's not ideal either to just not call the constructor and can lead to errors. Let's revert it.
There is two options : Maybe when type guessing, we could raise an exception indicating that the Input class cannot be automatically instantiated cause it has required constructor arguments.
Or Maybe we could add a way to add constructor arguments with `@GQL\Arg`, something like:
```

class SuperInput {
    constructor(int $superLevel) {
        ....
    }

and use it like that: 

@GQL\Arg(type="SuperInput", constructor={2})
```

PR #737, opened 2020-08-16
Make the docker image more usable

bartv2:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | yes/no
| Deprecations? | yes/no
| Tests pass?   | yes/no
| Documented?   | yes/no
| Fixed tickets | #... 
| License       | MIT

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,6 +5,6 @@
 !/benchmarks/
 !/composer.json
 !/phpunit.xml.*
-!/phpstan.neon.*
+!/phpstan*.neon
 !/.php_cs.*
 !/phpbench.json
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ FROM scratch AS test_source
 COPY benchmarks/ benchmarks/
 COPY src/ /src/
 COPY tests/ /tests/
-COPY phpunit.xml.* phpstan.neon.* .php_cs.* phpbench.json /
+COPY phpunit.xml.* phpstan*.neon .php_cs.* phpbench.json /
 
 FROM alpine:3.9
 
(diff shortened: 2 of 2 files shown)


mathroc:
composer 2 also supports parallel fetching of the dependencies, maybe it’ll be easier for maintenance to just wait for a stable release of composer (or include the alpha version)

.dockerignore
```diff
@@ -5,6 +5,6 @@
 !/benchmarks/
 !/composer.json
 !/phpunit.xml.*
-!/phpstan.neon.*
+!/phpstan*.neon
```
mcg-web, line 8: why this change?

Dockerfile
```diff
@@ -7,7 +7,7 @@ FROM scratch AS test_source
 COPY benchmarks/ benchmarks/
 COPY src/ /src/
 COPY tests/ /tests/
-COPY phpunit.xml.* phpstan.neon.* .php_cs.* phpbench.json /
+COPY phpunit.xml.* phpstan*.neon .php_cs.* phpbench.json /
```
mcg-web, line 10: why this change?

.dockerignore
murtukov: @mcg-web I think because i renamed the `phpstan.neon.dist` long time ago

bartv2: Yes the name changed and the config wasn't included

bartv2:
composer install is slow, doing parallel downloads helps a lot. This was the solution that i also use in other projects.

.dockerignore
mcg-web: ok nice, thank you!

PR #749, opened 2020-08-19
Add attributes `targetTypeQuery` and `targetTypeMutation` on `@Provider`

Vincz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| License       | MIT

This PR adds attributes `targetTypeQuery` and `targetTypeMutation` on `@Provider`
It will be used as default `targetType` for `@Query` or `@Mutation` if it is not set.

### DIFF
--- a/docs/annotations/annotations-reference.md
+++ b/docs/annotations/annotations-reference.md
@@ -433,8 +433,8 @@ You can use `@Access` and/or `@IsPublic` on a provider class to add default acce
 Optional attributes:
 
 -   **prefix**: A prefix to apply to all field names from this provider
--   **targetTypeQuery**: The default GraphQL type(s) to attach the provider `@Query` to
--   **targetTypeMutation**: The default GraphQL type(s) to attach the provider `@Mutation` to
+-   **targetQueryTypes**: The default GraphQL type(s) to attach the provider `@Query` to
+-   **targetMutationTypes**: The default GraphQL type(s) to attach the provider `@Mutation` to
 
 ## @Query
 
--- a/docs/annotations/annotations-reference.md
+++ b/docs/annotations/annotations-reference.md
@@ -382,9 +382,9 @@ class SecretArea {
 
 This annotation applies on methods for classes tagged with the `@Provider` annotation. It indicates that the method on this class will resolve a Mutation field.  
 The corresponding GraphQL field is added to the GraphQL type(s) following the logic:
-- The type(s) specified in the `targetType` attribute of the `@Mutation` annotation if it's defined.  
+- The type(s) specified in the `targetTypes` attribute of the `@Mutation` annotation if it's defined.  
   or
-- The type(s) specified in the `targetTypeMutation` attribute of the `@Provider` annotation if it's defined.  
+- The type(s) specified in the `targetMutationTypes` attribute of the `@Provider` annotation if it's defined.  
   or
 - The root Query type of the default schema (defined in configuration at key `overblog_graphql.definitions.schema.mutation` or `overblog_graphql.definitions.schema.default.mutation`).  
   
@@ -392,7 +392,7 @@ The class exposing the mutation(s) must be declared as a [service](https://symfo
 
 Optional attributes:
 
--   **targetType** : The GraphQL type(s) to attach the field to. It must be a mutation. (by default, it'll be the root Mutation type of the default schema. see [Default Schema](../definitions/schema.md#default-schema)). You can specify one or multiple target types.
+-   **targetTypes** : The GraphQL type(s) to attach the field to. It must be a mutation. (by default, it'll be the root Mutation type of the default schema. see [Default Schema](../definitions/schema.md#default-schema)). You can specify one or multiple target types.
 
 Example:
 
@@ -440,9 +440,9 @@ Optional attributes:
 
 This annotation applies on methods for classes tagged with the `@Provider` annotation. It indicates that on this class a method will resolve a Query field.  
 The corresponding GraphQL field is added to the GraphQL type(s) following the logic:
-- The type(s) specified in the `targetType` attribute of the `@Query` annotation if it's defined.  
+- The type(s) specified in the `targetTypes` attribute of the `@Query` annotation if it's defined.  
   or
-- The type(s) specified in the `targetTypeQuery` attribute of the `@Provider` annotation if it's defined.  
+- The type(s) specified in the `targetQueryTypes` attribute of the `@Provider` annotation if it's defined.  
   or
 - The root Query type of the default schema (defined in configuration at key `overblog_graphql.definitions.schema.query` or `overblog_graphql.definitions.schema.default.query`).  
   
@@ -450,7 +450,7 @@ The class exposing the query(ies) must be declared as a [service](https://symfon
 
 Optional attributes:
 
--   **targetType** : The GraphQL type(s) to attach the field to (by default, it'll be the root Query type of the default schema. see [Default Schema](../definitions/schema.md#default-schema)). You can specify one or multiple target types.
+-   **targetTypes** : The GraphQL type(s) to attach the field to (by default, it'll be the root Query type of the default schema. see [Default Schema](../definitions/schema.md#default-schema)). You can specify one or multiple target types.
 
 Example:
 
--- a/docs/annotations/annotations-reference.md
+++ b/docs/annotations/annotations-reference.md
@@ -381,12 +381,12 @@ class SecretArea {
 ## @Mutation
 
 This annotation applies on methods for classes tagged with the `@Provider` annotation. It indicates that the method on this class will resolve a Mutation field.  
-The corresponding GraphQL field is added to the GraphQL type(s) following the logic :
-- The type(s) specified in the `targetType` attribute of the `@Mutation` annotation if it defined.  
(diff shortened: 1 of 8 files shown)


src/Annotation/Provider.php
```diff
@@ -14,8 +14,20 @@ final class Provider implements Annotation
 {
     /**
      * Optionnal prefix for provider fields.
-     *
-     * @var string
```
murtukov, line 18: Any reason you removed the `@var` annotation?

docs/annotations/annotations-reference.md
```diff
@@ -427,11 +433,19 @@ You can use `@Access` and/or `@IsPublic` on a provider class to add default acce
 Optional attributes:
 
 -   **prefix** : A prefix to apply to all field names from this provider
+-   **targetTypeQuery** : The default GraphQL type(s) to attach the provider `@Query` to
+-   **targetTypeMutation** : The default GraphQL type(s) to attach the provider `@Mutation` to
```
murtukov, line 437:
```suggestion
-   **prefix**: A prefix to apply to all field names from this provider
-   **targetTypeQuery**: The default GraphQL type(s) to attach the provider `@Query` to
-   **targetTypeMutation**: The default GraphQL type(s) to attach the provider `@Mutation` to
```

docs/annotations/annotations-reference.md
```diff
@@ -381,7 +381,13 @@ class SecretArea {
 ## @Mutation
 
 This annotation applies on methods for classes tagged with the `@Provider` annotation. It indicates that the method on this class will resolve a Mutation field.  
-The resulting field is added to the root Mutation type (defined in configuration at key `overblog_graphql.definitions.schema.mutation`).  
+The corresponding GraphQL field is added to the GraphQL type(s) following the logic :
+- The type(s) specified in the `targetType` attribute of the `@Mutation` annotation if it defined.  
+  or
+- The type(s) specified in the `targetTypeMutation` attribute of the `@Provider` annotation if it is defined.  
+  or
+- The root Query type of the default schema (define in configuration at key `overblog_graphql.definitions.schema.mutation` or `overblog_graphql.definitions.schema.default.mutation`).
```
murtukov, line 389:
```suggestion
The corresponding GraphQL field is added to the GraphQL type(s) following the logic:
- The type(s) specified in the `targetType` attribute of the `@Mutation` annotation if it's defined.  
  or
- The type(s) specified in the `targetTypeMutation` attribute of the `@Provider` annotation if it's defined.  
  or
- The root Query type of the default schema (defined in configuration at key `overblog_graphql.definitions.schema.mutation` or `overblog_graphql.definitions.schema.default.mutation`).  
```

docs/annotations/annotations-reference.md
```diff
@@ -427,11 +433,19 @@ You can use `@Access` and/or `@IsPublic` on a provider class to add default acce
...
 ## @Query
 
 This annotation applies on methods for classes tagged with the `@Provider` annotation. It indicates that on this class a method will resolve a Query field.  
-By default, the resulting field is added to the root Query type (define in configuration at key `overblog_graphql.definitions.schema.query`).  
+The corresponding GraphQL field is added to the GraphQL type(s) following the logic :
+- The type(s) specified in the `targetType` attribute of the `@Query` annotation if it defined.  
+  or
+- The type(s) specified in the `targetTypeQuery` attribute of the `@Provider` annotation if it is defined.  
+  or
+- The root Query type of the default schema (define in configuration at key `overblog_graphql.definitions.schema.query` or `overblog_graphql.definitions.schema.default.query`).
```
murtukov, line 447:
```suggestion
The corresponding GraphQL field is added to the GraphQL type(s) following the logic:
- The type(s) specified in the `targetType` attribute of the `@Query` annotation if it's defined.  
  or
- The type(s) specified in the `targetTypeQuery` attribute of the `@Provider` annotation if it's defined.  
  or
- The root Query type of the default schema (defined in configuration at key `overblog_graphql.definitions.schema.query` or `overblog_graphql.definitions.schema.default.query`).  
```

src/Annotation/Provider.php
```diff
@@ -14,8 +14,20 @@ final class Provider implements Annotation
...
      */
     public string $prefix;
+
+    /**
+     * The default target types to attach the provider queries to.
+     *
+     * @var array<string>
+     */
+    public array $targetTypeQuery;
```
murtukov, line 25: Wouldn't it be better to name the property `$targetQueryTypes`?

src/Annotation/Provider.php
```diff
@@ -14,8 +14,20 @@ final class Provider implements Annotation
...
+
+    /**
+     * The default target types to attach the provider mutations to.
+     *
+     * @var array<string>
+     */
+    public array $targetTypeMutation;
```
murtukov, line 32: Wouldn't it be better to name the property `$targetMutationTypes`?

src/Config/Parser/AnnotationParser.php
```diff
@@ -712,6 +712,15 @@ private static function getGraphQLFieldsFromProviders(GraphClass $graphClass, st
 
                 $annotationTargets = $annotation->targetType ?? null;
 
+                if (null === $annotationTargets) {
+                    if ($annotation instanceof GQL\Mutation && isset($providerAnnotation->targetTypeMutation)) {
+                        $annotationTargets = $providerAnnotation->targetTypeMutation;
+                    }
+                    if ($annotation instanceof GQL\Query && isset($providerAnnotation->targetTypeQuery)) {
+                        $annotationTargets = $providerAnnotation->targetTypeQuery;
+                    }
```
murtukov, line 721:
```suggestion
                    if ($annotation instanceof GQL\Mutation && isset($providerAnnotation->targetTypeMutation)) {
                        $annotationTargets = $providerAnnotation->targetTypeMutation;
                    } elseif ($annotation instanceof GQL\Query && isset($providerAnnotation->targetTypeQuery)) {
                        $annotationTargets = $providerAnnotation->targetTypeQuery;
                    }
```

murtukov (changes requested):

src/Annotation/Provider.php
Vincz: You are right, but I wanted to stay consistent with the `targetType` of the `@Query` and let people know that `targetTypeQuery` and `targetType` are about the same feature.
Also, now it supports multiple types, so `targetTypes` would be better but it wouldn't be BC.

murtukov: You can add all names and mark old ones deprecated, so that it stays BC. It just requires a little more logic to check both fields and a comment, e.g:
```PHP
// TODO: Remove old property check in 1.1
$targetTypes = $anno->targetTypes ?? $anno->targetType ?? null;
```

src/Annotation/Provider.php
Vincz: Done

src/Annotation/Provider.php
Vincz: So, I marked the old one as deprecated and handle both cases as you suggest (and left one old `targetType` in the tests for now).
Should I also update the `UPGRADE-1.0.md` ?

docs/annotations/annotations-reference.md
```diff
@@ -426,17 +432,25 @@ You can use `@Access` and/or `@IsPublic` on a provider class to add default acce
 
 Optional attributes:
 
--   **prefix** : A prefix to apply to all field names from this provider
+-   **prefix**: A prefix to apply to all field names from this provider
+-   **targetTypeQuery**: The default GraphQL type(s) to attach the provider `@Query` to
+-   **targetTypeMutation**: The default GraphQL type(s) to attach the provider `@Mutation` to
```
murtukov, line 437:
```suggestion
-   **targetQueryTypes**: The default GraphQL type(s) to attach the provider `@Query` to
-   **targetMutytionTypes**: The default GraphQL type(s) to attach the provider `@Mutation` to
```

src/Annotation/Query.php
```diff
@@ -12,10 +12,16 @@
  */
 final class Query extends Field
 {
+    /**
+     * @deprecated
+     * @var array<string>
```
murtukov, line 17:
```suggestion
     * @var array<string>
     *
     * @deprecated This property is deprecated since 1.0 and will be removed in 1.1. Use $targetTypes instead.
```

src/Annotation/Provider.php
murtukov: I think it will be required in `UPGRADE-1.1.md` when the property is actually removed

Vincz: As always, thank you @murtukov  for the great review!

murtukov (approved):

src/Annotation/Mutation.php
```diff
@@ -12,10 +12,16 @@
  */
 final class Mutation extends Field
 {
+    /**
+     * @deprecated
+     * @var array<string>
```
murtukov, line 17:
```suggestion
     * @var array<string>
     *
     * @deprecated This property is deprecated since 1.0 and will be removed in 1.1. Use $targetTypes instead.
```

murtukov (changes requested):

murtukov (approved):

PR #751, opened 2020-08-20
Make it possible for scalar annotation to register Doctrine and PHP types to map

bartv2:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | not yet
| License       | MIT

When creating a scalar class it should be easy to also tell which types always map to this scalar. When the idea is accepted i will also update the documentation.

### DIFF
--- a/tests/Config/Parser/fixtures/annotations/Type/CustomScalarType.php
+++ b/tests/Config/Parser/fixtures/annotations/Type/CustomScalarType.php
@@ -1,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Overblog\GraphQLBundle\Tests\Config\Parser\fixtures\annotations\Type;
+
+use DateTime;
+use Doctrine\ORM\Mapping as ORM;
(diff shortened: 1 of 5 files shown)


tests/Config/Parser/fixtures/annotations/Type/CustomScalarType.php
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * @GQL\Type
+ */
+class CustomScalarType
+{
+    /**
+     * @ORM\Column(type="datetime", nullable=true)
+     * @GQL\Field
+     */
+    // @phpstan-ignore-next-line
```
murtukov, line 20:
```suggestion
    /**
     * @ORM\Column(type="datetime", nullable=true)
     * @GQL\Field
     * @phpstan-ignore-next-line
     */
```

PR #761, opened 2020-10-07
Update src/Generator/TypeBuilder.php

koekaverna:
Change self:: to static:: when refreing to class constant

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | #... 
| License       | MIT

In my project I need to extends `Overblog\GraphQLBundle\Generator\TypeBuilder` to overide `protected const EXTENDS;`
I use custom Enum in my models and I need to overide `EnumType` serrializer to make it works. So I updated `TypeBuilder` to use `static::EXTENDS` instead `self::EXTENDS`

### DIFF
--- a/src/Generator/TypeBuilder.php
+++ b/src/Generator/TypeBuilder.php
@@ -102,10 +102,10 @@ class TypeBuilder
 
         $class = $this->file->createClass($config['class_name'])
             ->setFinal()
-            ->setExtends(self::EXTENDS[$type])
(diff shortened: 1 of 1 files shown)


murtukov (approved):

koekaverna:
@murtukov thx!

PR #762, opened 2020-10-08
Fix Symfony setDeprecated() arguments

koekaverna:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | #... 
| License       | MIT

Since symfony/config 5.1: The signature of method "Symfony\Component\Config\Definition\Builder\NodeDefinition::setDeprecated()" requires 3 arguments: "string $package, string $version, string $message", not defining them is deprecated.

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -14,6 +14,7 @@ use Overblog\GraphQLBundle\EventListener\ErrorLoggerListener;
 use Overblog\GraphQLBundle\Executor\Executor;
 use Overblog\GraphQLBundle\ExpressionLanguage\ExpressionLanguage;
 use Overblog\GraphQLBundle\Resolver\FieldResolver;
+use Symfony\Component\Config\Definition\BaseNode;
 use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
 use Symfony\Component\Config\Definition\Builder\EnumNodeDefinition;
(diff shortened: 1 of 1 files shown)


murtukov:
@koekaverna I already fixed this in #715 and it was merged into 0.13. It just needs to be applied to master too.

@mcg-web this [PR](https://github.com/overblog/GraphQLBundle/pull/715/files) should also be applied to **master**

koekaverna:
@murtukov Okay, do it in your way.

koekaverna:
@murtukov 
when are you planning to do it?

murtukov:
@koekaverna the deprecation message says:

> The "%path%.%node%" configuration is deprecated since version 0.13 and will be removed in 1.0

The next major version is 1.0 (current master), so we don't need this BC-layer anymore, but need to remove the 'resolver_maps' option instead (as stated in the message). I guess this was the reason I didn't merge it into master.

@mcg-web can we remove this deprecated option in the **master**?

https://github.com/overblog/GraphQLBundle/blob/6192522c8376076d46e5dbe485d9d1cdc75e2c2d/src/DependencyInjection/Configuration.php#L223-L227

mcg-web:
@murtukov since next major version is the 1.0 we can safely make this obsolete and remove the configuration.

murtukov:
@koekaverna see #763

PR #770, opened 2020-10-30
Refactor TypeBuilder and resolve some todos

murtukov:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

- Add extended PHPDocs with code examples to each method of TypeBuilder
- Disable PHP-CS-Fixer rule to add unnecessary dots (they were added even after `:`
- Create static method `ExpressionLanguage::isStringWithTrigger`
- Other minor changes

### DIFF
--- a/src/Generator/TypeBuilder.php
+++ b/src/Generator/TypeBuilder.php
@@ -208,7 +208,7 @@ class TypeBuilder
      *
      *      fn() => [
      *          'name' => self::NAME,
-     *          'description' => 'Root query tyoe',
+     *          'description' => 'Root query type',
      *          'fields' => fn() => [
      *              'posts' => {@see buildField},
      *              'users' => {@see buildField},
@@ -537,7 +537,7 @@ class TypeBuilder
      *          'constraints' => {@see buildConstraints}
      *      ]
      *
-     * If only constraints provided, build {@see buildConstraints} directly.
+     * If only constraints provided, uses {@see buildConstraints} directly.
      *
      * @param array{
      *     constraints: array,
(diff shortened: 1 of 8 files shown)


mcg-web (approved):

src/Generator/TypeBuilder.php
```diff
@@ -293,35 +397,87 @@ protected function buildScalarCallback($callback, string $fieldName)
...
+                ->addArguments('value', 'args', 'context', 'info')
+                ->bindVar(TypeGenerator::GLOBAL_VARS);
 
-        // TODO (murtukov): replace usage of converter with ExpressionLanguage static method
-        if ($this->expressionConverter->check($resolve)) {
-            $injectErrors = ExpressionLanguage::expressionContainsVar('errors', $resolve);
+            $injectErrors = EL::expressionContainsVar('errors', $resolve);
 
             if ($injectErrors) {
                 $closure->append('$errors = ', Instance::new(ResolveErrors::class));
```
akomm, line 455: Having no **validation config**, but **errors** in the Expression might be a sign of an error. Still passing an **ResolveErrors** object might hide this fact. If you pass nothing, you would probably realize this via type or missing arg error. I'm not very tied to this, but just for you to consider if you did not already. :)

murtukov: @akomm Didn't quite understand.

> Still passing an ResolveErrors object might hide this fact

What fact can it hide?

akomm: *Having no validation config, but errors in the Expression might be a sign of an error.  Still passing an ResolveErrors object might hide this fact*

PR #772, opened 2020-11-03
Optimize config processing

murtukov:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no

Config processors are runtime processors, that change the config of types during the initialization.
There are currently 3 config processors:

- **AclConfigProcessor** - checks permissions to fields and if permission is not granted, it replaces the resolver with the following one:
  ```php
  function (): void {
      throw new UserWarning('Access denied to this field.');
  }
  ```
- **PublicFieldsFilterConfigProcessor** - this processor dynamically removes fields if the `public` callback returns false
- **WrapArgumentConfigProcessor** - this processor wraps ALL resolver closures into another closure only to wrap the $args param into `Argument` object.

Although these processor could also be optimized (they don't have to be executed always), this PR is focused on other things. 

If we look at any generated type (in the branch `master`), we can see that configs are wrapped into an arrow function:
```php
final class AddressType extends InputObjectType implements GeneratedTypeInterface
{
    public const NAME = 'Address';
    
    public function __construct(ConfigProcessor $configProcessor, GlobalVariables $globalVariables = null)
    {
        $configLoader = fn() => [
            'name' => self::NAME,
            'fields' => fn() => [
                // ...
            ],
        ];
        $config = $configProcessor->process(LazyConfig::create($configLoader, $globalVariables))->load();
        parent::__construct($config);
    }
}
```
The `webonyx/graphql-php` library doesn't accept closures as config, but arrays. In the code above the `$configLoader` is passed to `LazyConfig` where it is unwrapped by calling:
```php
$config = $configLoader($globalVariables);
```
`LazyConfig` is a simple class, that just takes a config closure, unwraps it, applies all 3 processors to it and returns config back as array. I don't know why config is wrapped into a closure, maybe for some historical reasons, because in previous versions it was used to pass `$globalVariables` to it, although we could just bind it with `use`:
```php
final class AddressType extends InputObjectType implements GeneratedTypeInterface
{
    public const NAME = 'Address';
    
    public function __construct(ConfigProcessor $configProcessor, GlobalVariables $globalVariables = null)
    {
        $configLoader = function ($globalVariables) {
            return [
                'name' => self::NAME,
                'fields' => function() use ($globalVariales) { 
                    return [
                        // ...
                    ],
                }
            ];
        };
        $config = $configProcessor->process(LazyConfig::create($configLoader, $globalVariables))->load();
        parent::__construct($config);
    }
}
```
Whatever the reason was, wrapping configs into a closure and passing `$globalVariables` into `LazyConfig` is an unnecessary overhead, therefore the generated classes can look like this:
```php
final class AddressType extends InputObjectType implements GeneratedTypeInterface
{
    public const NAME = 'Address';
    
    public function __construct(ConfigProcessor $configProcessor, GlobalVariables $globalVariables = null)
    {
        $config = [
            'name' => self::NAME,
            'fields' => fn() => [
                // ...
            ],
        ];
        $config = $configProcessor->process(LazyConfig::create($config))->load();
        parent::__construct($config);
    }
}
```
Furthermore the instantiation of the `LazyConfig` object doesn't have to be done inside the type's constructor, but inside the $configProcessor:
```php
final class AddressType extends InputObjectType implements GeneratedTypeInterface
{
    public const NAME = 'Address';
    
    public function __construct(ConfigProcessor $configProcessor, GlobalVariables $globalVariables = null)
    {
        $config = [
            'name' => self::NAME,
            'fields' => fn() => [
                // ...
            ],
        ];
        parent::__construct($configProcessor->process($config));
    }
}
```
And after all these changes, if you look at the `LazyConfig` class, you will see that its task could be moved directly into `ConfigProcessor`. Therefore the `LazyConfig` class is not needed anymore and can be removed.

This PR removes the `LazyConfig` class and reduces the number of function calls by type instantiations.

Note: don't be confused with the name `LazyConfig` because it doesn't really mean [lazy loading](https://en.wikipedia.org/wiki/Lazy_loading)


src/Definition/ConfigProcessor.php
```diff
@@ -6,7 +6,7 @@
 
 use Overblog\GraphQLBundle\Definition\ConfigProcessor\ConfigProcessorInterface;
 
-final class ConfigProcessor implements ConfigProcessorInterface
```
murtukov, line 9: ConfigProcessor don't have to implement ConfigProcessorInterface, because this class is a container for the **real** config processors. Only the 3 real config processors should implement this interface.

src/Definition/ConfigProcessor/AclConfigProcessor.php
```diff
@@ -47,21 +47,15 @@ public static function acl(array $fields, AccessResolver $accessResolver, callab
         return $fields;
     }
 
-    public function process(LazyConfig $lazyConfig): LazyConfig
+    public function process(array &$config): void
```
murtukov, line 50: Passing config array by reference to avoid copy of an array for performance reasons.

src/Definition/ConfigProcessor/ConfigProcessorInterface.php
```diff
@@ -4,9 +4,9 @@
 
 namespace Overblog\GraphQLBundle\Definition\ConfigProcessor;
 
-use Overblog\GraphQLBundle\Definition\LazyConfig;
+use Closure;
 
 interface ConfigProcessorInterface
 {
-    public function process(LazyConfig $lazyConfig): LazyConfig;
+    public function process(array &$config): void;
```
murtukov, line 11: The processors don't return the `LazyConfig` anymore, because:

- LazyConfig is removed
- The returned value is never used anyway

mcg-web (approved):

Issue #286, opened 2018-01-31
Pagination helper - backward count does not work with promises


akomm:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no
| Version/Branch   | dev-master

Paginator $count callback does work with promise when using forward, but not backward. In case of backward, you get:

`Object of class GraphQL\\Executor\\Promise\\Promise could not be converted to int`

mcg-web:
Try this please:

```php
$paginator = new Paginator(function($offset, $limit) {
    return $this->someLoader->load(new SomeKey($criteria, $offset, $limit))->then(function(array $ids) {
        return $this->someOtherLoader->loadMany($ids);
    });
}, Paginator::MODE_PROMISE);
```

akomm:
Sorry, I did it already, forgot to add it to the sample code.

akomm:
I think the reason, why the notice does not trigger in case of forward is because it is not accessed there, while the backward does access the Connection::$totalCount and therefore triggers the error. So probably the issue is in general, that the $count callback in Paginator is not supporting promise itself?

PR #778, opened 2020-11-06
Made it possible for totalCount to be a promise

rsolyanik:
An error occurs when trying to return a promise to totalCount
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | #286 
| License       | MIT

### DIFF
--- a/src/Relay/Connection/ConnectionInterface.php
+++ b/src/Relay/Connection/ConnectionInterface.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace Overblog\GraphQLBundle\Relay\Connection;
 
+use GraphQL\Executor\Promise\Promise;
+
 interface ConnectionInterface
 {
     /**
@@ -33,14 +35,16 @@ interface ConnectionInterface
     public function setPageInfo(PageInfoInterface $pageInfo): void;
 
     /**
-     * Get the total count.
+     * Get the total count or promise that returns the total count.
      *
-     * @return int
+     * @return int|Promise|null
      */
-    public function getTotalCount(): ?int;
+    public function getTotalCount();
 
     /**
-     * Set the total count.
+     * Set the total count or promise that returns the total count.
+     *
+     * @param int|Promise $totalCount
      */
-    public function setTotalCount(int $totalCount): void;
+    public function setTotalCount($totalCount): void;
 }
--- a/src/Relay/Connection/Output/Connection.php
+++ b/src/Relay/Connection/Output/Connection.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Overblog\GraphQLBundle\Relay\Connection\Output;
 
+use GraphQL\Executor\Promise\Promise;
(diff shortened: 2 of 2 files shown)


src/Relay/Connection/ConnectionInterface.php
```diff
@@ -35,12 +35,14 @@ public function setPageInfo(PageInfoInterface $pageInfo): void;
     /**
      * Get the total count.
      *
-     * @return int
+     * @return int|object|null
      */
```
murtukov, line 39: Need to specify the exact returned object class or interface (promise in this case, I guess). And updating the description would be nice too.

src/Relay/Connection/ConnectionInterface.php
```diff
@@ -35,12 +35,14 @@ public function setPageInfo(PageInfoInterface $pageInfo): void;
...
-    public function getTotalCount(): ?int;
+    public function getTotalCount();
 
     /**
      * Set the total count.
+     *
+     * @param int|object $totalCount
```
murtukov, line 45: Same here

src/Relay/Connection/Output/Connection.php
```diff
@@ -16,7 +16,9 @@ class Connection implements ConnectionInterface
     protected array $edges;
 
     protected ?PageInfoInterface $pageInfo;
-    protected ?int $totalCount = null;
+
+    /** @var int|object|null */
+    protected $totalCount = null;
```
murtukov, line 21: Same here

src/Relay/Connection/Output/Connection.php
```diff
@@ -16,7 +17,9 @@ class Connection implements ConnectionInterface
     protected array $edges;
 
     protected ?PageInfoInterface $pageInfo;
-    protected ?int $totalCount = null;
+
+    /** @var int|Promise|null Total count or promise that returns the total count */
```
murtukov, line 21:
```suggestion
    /** 
     * Total count or promise that returns the total count.
     * @var int|Promise|null
     */
```

murtukov:
ping @mcg-web 

Could you review this PR? I don't work with Relay, so I can't say if it's a good thing to allow to return a promise from the getter.

murtukov:
@rsolyanik before pushing any changes you have to execute `composer fix-cs` to bring the code to predefined coding styles and `composer static-analysis` to check if there are any problems. If there are any problems, Travis-CI check won't pass.

rsolyanik:
> ping @mcg-web
> 
> Could you review this PR? I don't work with Relay, so I can't say if it's a good thing to allow to return a promise from the getter.
This is necessary to load total counts via data loader in nested paged lists
Something like 
`{
  customers(first: 1000) {
    edges {
      node {
        id,
        loyaltyAccounts {
          totalCount
          edges {
            node {
              ....
            }
          }
        }
      }
    }
  }
}`

murtukov (approved):

PR #782, opened 2020-11-11
Made it possible for totalCount to be a promise

rsolyanik:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| License       | MIT

### DIFF
--- a/src/Relay/Connection/ConnectionInterface.php
+++ b/src/Relay/Connection/ConnectionInterface.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Overblog\GraphQLBundle\Relay\Connection;
 
+use GraphQL\Executor\Promise\Promise;
 use Overblog\GraphQLBundle\Relay\Connection\Output\PageInfo;
 
 interface ConnectionInterface
@@ -37,16 +38,16 @@ interface ConnectionInterface
     public function setPageInfo(PageInfoInterface $pageInfo);
 
     /**
-     * Get the total count.
+     * Get the total count or promise that returns the total count.
      *
-     * @return int
+     * @return int|Promise
      */
-    public function getTotalCount(): ?int;
+    public function getTotalCount();
 
     /**
-     * Set the total count.
+     * Set the total count or promise that returns the total count.
      *
-     * @param int $totalCount
+     * @param int|Promise $totalCount
      */
-    public function setTotalCount(int $totalCount);
+    public function setTotalCount($totalCount);
 }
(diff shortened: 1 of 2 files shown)


mcg-web:
Sorry but all released versions are feature freeze. We only accept bugfix or security fix right now for these branches. Your feature add bc so can't definitely not be accepted.

rsolyanik:
> Sorry but all released versions are feature freeze. We only accept bugfix or security fix right now for these branches. Your feature add bc so can't definitely not be accepted.

@mcg-web this don't break bc, but fixes error `Object of class GraphQL\\Executor\\Promise\\Promise could not be converted to int` when trying load total counts in nested lists via data loaders for such queries as` { customers(first: 1000) { edges { node { id, loyaltyAccounts { totalCount edges { node { .... } } } } } } }
`

mcg-web:
Changing the signature of an interface is a bc. If the fix introduce any bc it can be allowed on this branch maybe it can be fixed a different way?

murtukov:
@mcg-web Well, that's a tricky case. What if fixing a bug requires changing the signature of a function? Should we now just leave bugs only because of it? And is it a new feature anyway? I thinks it was supposed to be able to return promises from the beginning and if doesn't work as supposed, then it's a bug.

murtukov:
@mcg-web Well, that's a tricky case. What if fixing a bug requires changing the signature of a function? Should we now just leave bugs only because of it? And is it a new feature anyway? I thinks it was supposed to be able to return promises from the beginning and if doesn't work as supposed, then it's a bug.

What is a BC break? It's something that makes your code fail. The following signature:
```php
public function getTotalCount()
```
is a superset of:
```php
public function getTotalCount(): ?int
```
How can it possibly make your code fail?

mcg-web:
@murtukov there's 2 methods changed, the getter doesn't effectively break anything but what about the setter ?

murtukov:
@mcg-web the same with the setter, he just removed the type hint. I mean how can it possibly break anything?

mcg-web:
This is a public interface. It scope doesn't stop to the bundle, so yes this interface was implemented by other users since it was introduced to help doing that. Changing a public method of a non internal class or interface is a BC.

murtukov:
@mcg-web Then we return back to the previous questions:

1. Is it a bug? Because the documentation says
> The bundle is totally "promise ready"

   yet it fails to return a promise

2. If yes, then what is more important, fixing the bug or stay BC?

mcg-web:
There is always a way of fixing this without breaking BC (some dedicated getter and setter to deal with promise?) , that what I say higher up. I'm not arguing about fixing this bug but the "how" does not follow the release policies of this bundle. We only allow break BC for security reasons. The contract is while updating a minors or a patch release the bundle should not break anything.

Issue #775, opened 2020-11-06
[RFC] Refactor GlobalVariables and resolver expressions


murtukov:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | no
| BC Break report? | no
| RFC?             | yes
| Version/Branch   | 1.0 (master)

Some of the proposed changes are BC-break.

#### 1. Rename `GlobalVariables` class.

`GlobalVariables` is a service passed to every generated GraphQL type. It contains all necessary services and references to other types to configure a GraphQL schema. The name `GlobalVariables` doesn't really explain anything, I suggest to rename it to better reflect it's purpose.

Suggested name: `GraphQLServices`
Other possible names: `SchemaServices`, `TypeServices`, `TypeHelper`

#### 2. Move `ConfigProcessor` into `GlobalVariables`

Every generated type also gets `ConfigProcessor` object, which is only used in types and nowhere else. If `GlobalVariables` contains all necessary services for types, then why not also moving the `ConfigProcessor` inside it?

Before:
```php
final class QueryType extends ObjectType implements GeneratedTypeInterface
{
    public const NAME = 'Query';
    
    public function __construct(ConfigProcessor $configProcessor, GraphQLServices $services)
    {
        $config = [
            'name' => self::NAME,
            'fields' => fn() => [
                'post' => [
                    'type' => Type::string(),
                    'resolve' => fn() => "Hello, World!",
                ],
            ],
        ];

        parent::__construct($configProcessor->process($config));
    }
}
```

After:
```php
final class QueryType extends ObjectType implements GeneratedTypeInterface
{
    public const NAME = 'Query';
    
    public function __construct(GraphQLServices $services)
    {
        $config = [
            'name' => self::NAME,
            'fields' => fn() => [
                'post' => [
                    'type' => Type::string(),
                    'resolve' => fn() => "Hello, World!",
                ],
            ],
        ];

        parent::__construct($services->get('configProcessor')->process($config));
    }
}
```

#### 3. Add shortcuts into `GraphQLServices`

Add methods for frequently used services, like `typeResolver`, `resolverResolver` and `mutationResolver`

Before:
```php
final class QueryType extends ObjectType implements GeneratedTypeInterface
{
    public const NAME = 'Query';
    
    public function __construct(GraphQLServices $services)
    {
        $config = [
            'name' => self::NAME,
            'fields' => fn() => [
                'users' => [
                    'type' => $services->get('typeResolver')->resolve('User'),
[...]

PR #786, opened 2020-11-30
Rename classes, tags and functions

murtukov:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | #775 

Implementation of #775

List of renames:

- Class `GlobalVariables` -> `GraphQLServices`
- Class `ResolverResolver` -> `QueryResolver`
- Interface `ResolverInterface` -> `QueryInterface`
- Service tag `overblog_graphql.global_variable` -> `overblog_graphql.service`
- Service tag `overblog_graphql.resolver` -> `overblog_graphql.query`
- Expression function `resolver` -> `query`

Also all UPGRADE-files for previous versions have been removed.


UPGRADE-1.0.md
```diff
@@ -4,8 +4,10 @@ UPGRADE FROM 0.13 to 1.0
 # Table of Contents
 
 - [Customize the cursor encoder of the edges of a connection](#customize-the-cursor-encoder-of-the-edges-of-a-connection)
-- [Change arguments of `TypeGenerator`](#change-arguments-of-typegenerator)
+- [Change arguments of `TypeGenerator`](#change-arguments-of-typegenerator-class)
 - [Add magic `__get` method to `ArgumentInterface` implementors](#add-magic-__get-method-to-argumentinterface-implementors)
+- [Rename `GlobalVariables` to `GraphQLServices`](#)
```
mcg-web, line 9: missing links

UPGRADE-1.0.md
```diff
@@ -4,8 +4,10 @@ UPGRADE FROM 0.13 to 1.0
...
+- [Change `overblog_graphql.global_variable` tag]()
```
mcg-web, line 11: Missing parts for resolver, mutation, all public tags, interfaces, classes ... changes (just a reminder)

docs/definitions/expression-language.md
```diff
@@ -320,11 +320,11 @@ Examples
 @=getUser().firstName === 'adam'
 ```
 
+
 ## Registered variables:
 
 | Variable             | Description  | Scope|
 |:-------------------- |:------------ |:---- |
-| `typeResolver`       | An object of class `Overblog\GraphQLBundle\Resolver\TypeResolver`| global|
```
mcg-web, line 327: why this change ?

mcg-web (changes requested):
All these changes make sense but also make the migration from `0.13` to `1.0` a very heavy task. We should maybe introduce all these in a `0.14` version with deprecations so it ease the process.
What do you think of this ?

murtukov:
@mcg-web I think making 0.14 as a transition version is a good idea.

docs/definitions/expression-language.md
murtukov: @mcg-web It's a mistake. First I removed `typeResolver` from expressions, then I rerolled this change, but forgot to reroll changes done to the documentation. I'll fix this.

UPGRADE-1.0.md
murtukov: @mcg-web same here, I'll write the UPDATE documentation.

UPGRADE-1.0.md
murtukov: @mcg-web Yep, I will write the UPDATE documentation.

murtukov:
@mcg-web I made it BC. Also I removed all UPGRADE files for previous versions. We can have a single UPGRADE.md file in each branch, like `webonyx/graphql-php` does.

Vincz (approved):

src/Definition/Resolver/ResolverInterface.php
```diff
@@ -4,6 +4,10 @@
 
 namespace Overblog\GraphQLBundle\Definition\Resolver;
 
+/**
```
mcg-web, line 7: also add a `trigger_error` call please

src/DependencyInjection/Compiler/ResolverTaggedServiceMappingPass.php
```diff
@@ -8,6 +8,12 @@
 use function is_string;
 use function sprintf;
 
+/**
```
mcg-web, line 11: also add a `trigger_error` call please

src/ExpressionLanguage/ExpressionFunction/GraphQL/Mutation.php
```diff
@@ -8,11 +8,32 @@
...
+    public function __construct($name = self::NAME)
     {
         parent::__construct(
             $name,
-            fn ($alias, $args = '[]') => "$this->globalVars->get('mutationResolver')->resolve([$alias, $args])"
+            function (string $alias, ...$args) {
+                $count = count($args);
+
+                // TODO: remove the following if-else-block in 1.0
+                if (1 === $count && '$' !== $args[0][0]) {
```
mcg-web, line 21: `$args[0][0]` can be unset, so should we not test it to avoid notice ?

mcg-web (changes requested):
We should also keep some legacy tests.

src/Definition/Resolver/ResolverInterface.php
murtukov: How do I add `trigger_error` in an interface?

mcg-web: You should just added before the class or the interface definition.
You can see examples here:
https://symfony.com/doc/current/contributing/code/conventions.html

murtukov: Done

src/DependencyInjection/Compiler/ResolverTaggedServiceMappingPass.php
murtukov: done

src/ExpressionLanguage/ExpressionFunction/GraphQL/Mutation.php
murtukov: `$args[0][0]` can never be unset. The `1 === $count` makes sure that `$args[0]` is set. Furthermore all items in the `$args` array are always non-empty strings, which means `$args[0][0]` points to the first letter of the string and is never unset.

Now a little explanation, why I did this check. We are changing the signature of the `mutation` function.

old: `mutation(string $alias, array $args = [])`
new: `mutation(string $alias, ...$args)`

The` if-block` tries to guess what signature user currently tries to use. If he calls the function with only 1 argument, which is an array (`'[' === $args[0][0]`), then it considers it old signature call, otherwise it uses the new signature.

The only drawback here is the case, where user tries to use the new signature and sends 1 array-argument:
`mutation("my_resolver", ['name', 'lastname'])`
So the bundle will think it's and old signature and will try to send `name` and `lastname` as 2 separate arguments to the resolver.

murtukov:
@mcg-web I tweaked some some tests to use legacy code, so there are no `@codeCoverageIgnore` tags required anymore

src/ExpressionLanguage/ExpressionFunction/GraphQL/Mutation.php
mcg-web: yes I agree, I don't see any easy way to make the diffence between old and new signature in this case without using reflection.

murtukov: @mcg-web so what do you think I should do? Should I leave it as is?

murtukov:
@mcg-web 

I added some additional checks with reflection to decide, which signature should be used.

Here are some examples of which signature is used by certain calls:

1. `mutation('get_posts')` - **new signature**
2. `mutation('get_posts', 'hello')` - **new signature**
3. `mutation('get_posts', [args, 'test'])`:
    - If the array contains the same number of elements as the number of arguments in proxy resolver: **old signature**
    - Otherwise: **new signature**

mcg-web (approved):

PR #790, opened 2020-12-22
Fixes #789 - pass the full query to the executor

### LINKED ISSUE #789: OperationName is not passed to Events when using Batch query
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no
| Version/Branch   | 0.13 and below

We provide an operationName for most if not all of our queries, and recently added an event listener to look at what operations were being executed.

When we use the batch endpoint, the line here does not pass the operationName through:
https://github.com/overblog/GraphQLBundle/blob/2b954fd7adc206a97b1fc013c5209e1cbc8bbebc/src/Controller/GraphController.php#L101

As such, the operationName is always null.

I've manually modified our code and find passing the operationName here works fine.

edhgoose:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | n/a - I haven't been able to do this yet, will try to shortly
| Documented?   | no
| Fixed tickets | #789 
| License       | MIT

The operationName is not passed when using the batch endpoint. It seems the code as it is explicitly passes the query and the variables, and this fix instead passes all elements of the query through. I think this is simpler - but I'm concerned I may be missing something as to why they are explicitly passed. 

The alternative would be as follows:

```
->execute($schemaName, ['query' => $query['query'], 'variables' => $query['variables'], 'operationName' => $query['operationName'])
```

I would welcome feedback from someone who knows the code base better. Thank you.

### DIFF
--- a/src/Controller/GraphController.php
+++ b/src/Controller/GraphController.php
@@ -135,7 +135,7 @@ class GraphController
 
         foreach ($queries as $query) {
             $payload = $this->requestExecutor
-                ->execute($schemaName, ['query' => $query['query'], 'variables' => $query['variables']])
+                ->execute($schemaName, $query)
                 ->toArray();
             if (!$this->useApolloBatchingMethod) {
                 $payload = ['id' => $query['id'], 'payload' => $payload];


mcg-web:
Thanks for the fix @edhgoose ! Can you please provide the patch for 0.11 branch please ? Then we can apply it to versions above.

edhgoose:
Hey @mcg-web - thanks. I've rebased off v0.11.19 and changed the base branch to 0.11. Does that work for you?

mcg-web (approved):

mcg-web:
Thank you @edhgoose !

edhgoose:
Brilliant, thanks @mcg-web. I'm unfamiliar with your release process, do you know when this will be released?

mcg-web:
I'll release a new version for 0.11, 0.12 and 0.13 before Friday.

Issue #707, opened 2020-07-09
Improve annotations


murtukov:
@Vincz  I went through the documentation and gathered some notes. Here are couple of improvements that could be done:

### 1. Add support for type auto-guessing from property type-hints

e.g. this:
```php
class Hero {
    /**
     * @GQL\Field
     */
    public string $name;

    /**
     * @GQL\Field
     */
    public ?string $secret;
}
```
should be equal to this:
```php
class Hero {
    /**
     * @GQL\Field(type="String!")
     */
    public string $name;

    /**
     * @GQL\Field(type="String")
     */
    public ?string $secret;
}
```

### 2. Omit the attribute's key
Treat the first argument as "type" by default

e.g. this:
```php
    /**
     * @GQL\Field("Float!")
     */
    public $latitude;
```

should be equal to this:

```php
    /**
     * @GQL\Field(type="Float!")
     */
    public $latitude;
```

The `name` attribute could be default instead of `type` as well, we just need to think, what makes more sense.

The same should be applied for @Arg.

The Symfony's `@Route` uses such syntax, See: https://symfony.com/doc/current/routing.html

### 3. Detect the `@param` and `@return`annotations for auto-guessing
As you stated in the documentation, some types can be auto-guessed from the return type of the method, but if it's `array`, then you need to declare the type explicitly:

```php
/**
 * @GQL\Provider
 */
class UsersProviders {

    /**
     * @GQL\Query(type="[User]", name="users")
     */
    public function getUsers(): array {
        return $this->repository->findAll();
    }
}
```
We could omit explicit type declaration, e.g.:
```php
/**
 * @GQL\Provider
 */
class UsersProviders {

    /**
     * @GQL\Query(name="users")
     *
     * @return User[]
     */
    public function getUsers(): array {
        return $this->repository->findAll();
    }
}
```
The following syntaxes should be valid: 
| Tag                     | GraphQL type       |
| :--------------------- | :----------------------- |
| `User[]`               | `[Users!]!`             | 
| `User[]\|null`       | `[Users!]`               |
| `array<User>`    | `[Users!]!`             |
| `array<User\|null>`    | `[Users]!`       |
| `null\|array<User\|null>`    | `[Users]` |
| `null\|array<User>`    | `[Users!]` |
| * `iterable<User>`    | `[Users]` |

`iterable` should have all the same forms as `array`.

[...]

PR #801, opened 2021-01-02
Annotations refactoring & PHP 8 Attributes

Vincz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Documented?   | partly
| Fixed tickets | #707


murtukov:
I thought we are gonna discuss it in January first

Vincz:
> I thought we are gonna discuss it in January first

@murtukov  We will. I had a bit of free time and I just wanted to implement the PHP 8 attributes and try to clean the annotations code base (also based on #707). But everything is still open to discussion.

Vincz:
@mcg-web Hey man!
I have a couple of issues with php stan as you can see here : https://github.com/overblog/GraphQLBundle/pull/801/checks?check_run_id=1637669155

I have a bunch of missing type hints but I'm not sure how to handle them when a parameter can accept multiple types (string or array for example).
Also, there is a bunch of issue related to PHP 8 (PHP stan does not know some classes or methods specifics to PHP8). Not sure how to handle this as well.
Could you help me with this?

Vincz:
@mcg-web Another problem is with the test coverage. As the PHP 8 attribute parser tests only run with PHP8, the coverage drops when used with PHP 7.

mcg-web:
Hey @Vincz ! 

> I have a bunch of missing type hints but I'm not sure how to handle them when a parameter can accept multiple types (string or array for example).

you should use phpdoc right now since union is available only for php >= 8.0

```php
/**
* @param string|array $myArg
*/
function ($myArg) {}
```

> Another problem is with the test coverage. As the PHP 8 attribute parser tests only run with PHP8, the coverage drops when used with PHP 7.

we can try using php 8.0 for coverage tests

src/Config/Parser/AttributeParser.php
```diff
@@ -0,0 +1,32 @@
...
+        $attributes = [];
+
+        switch (true) {
+            case $reflector instanceof ReflectionClass:
+            case $reflector instanceof ReflectionMethod:
+            case $reflector instanceof ReflectionProperty:
+                $attributes = $reflector->getAttributes();
+        }
+
+        // @phpstan-ignore-line
```
murtukov, line 29:
```suggestion
        // @phpstan-ignore-next-line
```

src/Config/Parser/MetadataParser/ClassesTypesMap.php
```diff
@@ -0,0 +1,76 @@
...
+
+        return null;
+    }
+
+    /**
+     * Resolve the class name associated with given type
+     */
+    public function resolveClass(string $typeName): ?string
+    {
+        return isset($this->classesMap[$typeName]) ? $this->classesMap[$typeName]['class'] : null;
```
murtukov, line 50:
```suggestion
        return $this->classesMap[$typeName]['class'] ?? null;
```

Vincz:
Ok guys, this PR is almost ready.
I still need to update the documentation. 
Trying to make a chinese panda happy, all features from #707 have been implemented (except the 'get' removal from queries).
- The nested annotations have been deprecated in favor of a flat config.
- We now have a type guesser based on doc block `@var` and `@return`
- PHP8 attributes are supported.
- We have shortcuts on annotations (ex: `@Field('myfield')`) - I'm not satisfied with the implementation but I'm waiting for this PR `https://github.com/doctrine/annotations/pull/391` to be merged, so I can submit another one to handle default property on annotation, so we can use more explicit configuration property instead of `value`.

We would need to be able to run tests in a PHP8 environment to support attributes (at the moment tests related to attributes are skipped if not in PHP8).
Some config keys have been renamed for builders so it's a BC break (is that ok if we targeting v1.0 with this new features or should I provider Backward compatibility).

Let me know what you think!

src/Annotation/Annotation.php
```diff
@@ -4,6 +4,13 @@
...
 namespace Overblog\GraphQLBundle\Annotation;
 
-interface Annotation
+use Doctrine\Common\Annotations\AnnotationException;
+
+abstract class Annotation
 {
+    protected function cumulatedAttributesException(string $attribute, string $value, string $attributeValue): void
+    {
+        $annotationName = str_replace('Overblog\GraphQLBundle\Annotation\\', '', get_class($this));
```
murtukov, line 13:
```suggestion
        $annotationName = (new ReflectionClass($this))->getShortName();
```

src/Annotation/Annotation.php
```diff
@@ -4,6 +4,13 @@
 
 namespace Overblog\GraphQLBundle\Annotation;
 
-interface Annotation
+use Doctrine\Common\Annotations\AnnotationException;
```
murtukov, line 7:
```suggestion
use Doctrine\Common\Annotations\AnnotationException;
use ReflectionClass;
```

src/Annotation/Builder.php
```diff
@@ -0,0 +1,40 @@
...
+     * @var string
+     */
+    public string $value;
+
+    /**
+     * The builder config.
+     *
+     * @var mixed
+     */
+    public $config = [];
```
murtukov, line 33: According to the constructor, this property should always be `array`.

src/Annotation/Annotation.php
Vincz: I can't because of Relay Annotations with a different namespace (https://github.com/overblog/GraphQLBundle/blob/master/src/Annotation/Relay/Connection.php)

Vincz:
Hey @murtukov! I'm trying to run the test on Github with PHP 8 (see my latest commit). 
It failed because of the PHP Generator library. Do you know if it is compatible with PHP 8 and if so, would you mind to update the composer to allow it? 
Thanks man!

murtukov:
@Vincz released a new version of PHPCodeGenerator. Change the composer requirement to 0.1.5

Vincz:
> @Vincz just released a new version of PHPCodeGenerator. Change the composer requirement to 0.1.5

Thank you!

Vincz:
@murtukov It break the PHP 7 tests, because:

```
➜  ~ $ echo '<?php new flu(); ?>' | php
PHP Fatal error:  Uncaught Error: Class 'flu' not found in Standard input code:1
```
but...
```
➜  ~ $ echo '<?php new flu(); ?>' | php8
PHP Fatal error:  Uncaught Error: Class "flu" not found in Standard input code:1
```

docs/annotations/annotations-reference.md
```diff
@@ -38,12 +38,27 @@ class Coordinates {
 }
 ```
 
+## Annotations shortcuts
+
+When using annotations or attributes, you can use a shortened version of the syntax.  
+Example without:  
+`@GQL\Field(name="MyField")` 
+
+Exampel with shortcut:
```
murtukov, line 47:
```suggestion
Example with shortcut:  
```

docs/annotations/annotations-reference.md
```diff
@@ -38,12 +38,27 @@ class Coordinates {
...
+## Annotations shortcuts
+
+When using annotations or attributes, you can use a shortened version of the syntax.  
+Example without:  
+`@GQL\Field(name="MyField")` 
+
+Exampel with shortcut:  
+`@GQL\Field("MyField")`
+
+When using the shortened version, it is the attribute describe as `default attribute` in the following list that will be set if attribute name is not specified.
```
murtukov, line 50: This sentence is built wrong. I would fix it, but I just don't understand it 😅

docs/annotations/annotations-reference.md
```diff
@@ -110,11 +127,11 @@ class Hero {
 
 ## @Arg
 
-This annotation is used in the `args` attribute of a `@Field` or `@Query` or `@Mutation` to define an argument.
+This annotation is used in conjonction with a `@Field`, a `@Query` or `@Mutation` to define an argument.
```
murtukov, line 130:
```suggestion
This annotation is used in conjunction with a `@Field`, a `@Query` or `@Mutation` to define an argument.
```

I feel a French trace here 😏

docs/annotations/annotations-reference.md
```diff
@@ -131,23 +148,52 @@ Example:
...
+     * @GQL\Arg(name="nameStartsWith", type="String", description="Retrieve only heroes with name starting with")
      */
     public $friends;
 }
 ```
 
+## @ArgsBuilder
+
+This annotation is used in conjonction with a `@Field`, a `@Query` or `@Mutation` to generate the field arguments.
+It is used to set a arguments builder for a field (see [Args builders](../definitions/builders/args.md)))
```
murtukov, line 168:
```suggestion
This annotation is used in conjunction with a `@Field`, a `@Query` or `@Mutation` to generate the field arguments.
It is used to set an arguments builder for a field (see [Args builders](../definitions/builders/args.md)))
```

docs/annotations/index.md
```diff
@@ -116,6 +162,34 @@ In the previous example, the generated `resolve` config of the `something` field
 
 ## Type & Args auto-guessing
 
+When no explicit `type` is defined on a `@Field` (`@Query`, or `@Mutation`), the bundle will try to guess it using the following methods:
+- **Doc Block** The guess will be made from `@var` or `@return` in Doc block
```
murtukov, line 166:
```suggestion
- **DocBlock** The guess will be made from the `@var` and `@return` tags in DocBlock
```

docs/annotations/index.md
```diff
@@ -116,6 +162,34 @@ In the previous example, the generated `resolve` config of the `something` field
...
+- **Type hint** The guess will be made from type hint 
+- **Doctrine annotations** The guess will be made from doctrine annotations
+
+The system will try every method, in the above order and return a type as soon as one of the guessing method return one.
```
murtukov, line 170:
```suggestion
If the `type` option is not defined explicitly on the `@Field`, `@Query` or `@Mutation`, the bundle will try to guess it from other DocBlock annotations or from the PHP type-hint, in the following order:

1. `@var` and `@return` annotations
2. type-hint 
3. Doctrine annotations

It will stop on the first successful guess.
```

Just tried to simplify this. Maybe need some examples.

docs/annotations/index.md
```diff
@@ -116,6 +162,34 @@ In the previous example, the generated `resolve` config of the `something` field
...
+
+### @Field type auto-guessing from Dock Block
+
+The type of the `@Field` annotation can be auto-guessed if its Dock Block describe a known type. It is the more precise auto-guessing as it supports collection of object type.
```
murtukov, line 174:
```suggestion
### @Field type guessing from DocBlock

The `type` option of the `@Field` annotation can be guessed if its DocBlock describes a known type. It is a more precise guessing as it supports collections of objects, e.g. `Users[]` or `<Users>.
```

src/Annotation/Description.php
```diff
@@ -4,20 +4,29 @@
 
 namespace Overblog\GraphQLBundle\Annotation;
 
+use \Attribute;
+use Doctrine\Common\Annotations\NamedArgumentConstructorAnnotation;
+
 /**
- * Annotation for GraphQL description.
+ * Annotation for GraphQL to mark a field as deprecated.
```
murtukov, line 11: class Description to mark a field as deprecated??

src/Annotation/Arg.php
```diff
@@ -43,4 +47,18 @@ final class Arg implements Annotation
...
     public $default;
+
+    /**
+     * @param mixed|null $default 
+     */
+    public function __construct(string $name, string $type, ?string $description = null, $default = null, ?string $value = null)
+    {
+        if ($value && $name) {
+            $this->cumulatedAttributesException('name', $value, $name);
+        }
```
murtukov, line 58: Why checking `$name`, if its type-hint is non-nullable?

src/Annotation/Enum.php
```diff
@@ -4,23 +4,39 @@
...
+    public ?string $name;
 
     /**
      * @var array<\Overblog\GraphQLBundle\Annotation\EnumValue>
+     * 
+     * @deprecated
      */
     public array $values;
+
+    public function __construct(?string $name = null, array $values = [], ?string $value = null)
```
murtukov, line 34: Are you sure that `$name` is nullable?

src/Annotation/EnumValue.php
```diff
@@ -4,26 +4,40 @@
...
-    public string $description;
+    public ?string $description;
 
     /**
      * @var string
      */
-    public string $deprecationReason;
+    public ?string $deprecationReason;
+
+    public function __construct(?string $name = null, ?string $description = null, ?string $deprecationReason = null, ?string $value = null)
```
murtukov, line 34: Same here with `$name`

src/Annotation/Query.php
```diff
@@ -4,25 +4,46 @@
...
  */
+#[Attribute(Attribute::TARGET_METHOD)]
 final class Query extends Field
 {
     /**
-     * @var array<string>
+     * The target types to attach this query to.
      *
-     * @deprecated This property is deprecated since 1.0 and will be removed in 1.1. Use $targetTypes instead.
+     * @var array<string>
```
murtukov, line 22:
```suggestion
     * @var array<string>|null
```

src/Annotation/Relay/Connection.php
```diff
@@ -12,19 +15,26 @@
...
  */
-final class Connection extends Type
+#[Attribute(Attribute::TARGET_CLASS)]
+final class Connection extends Type implements NamedArgumentConstructorAnnotation
 {
     /**
      * Connection Edge type.
-     *
+     * 
      * @var string
```
murtukov, line 24:
```suggestion
     * @var string|null
```

src/Annotation/Relay/Connection.php
```diff
@@ -12,19 +15,26 @@
...
      */
-    public string $edge;
+    public ?string $edge;
 
     /**
      * Connection Node type.
-     *
+     * 
      * @var string
```
murtukov, line 31:
```suggestion
     * @var string|null
```

src/Annotation/Scalar.php
```diff
@@ -4,22 +4,35 @@
...
-
+    public ?string $name;
 
     /**
      * @var string
      */
-    public string $scalarType;
+    public ?string $scalarType;
+    
+    public function __construct(?string $name = null, ?string $scalarType = null, ?string $value = null)
```
murtukov, line 30:
```suggestion
    public function __construct(string $name = null, string $scalarType = null, string $value = null)
```
Just for the info: if an argument has a default value of null, then you don't have to explicitly declare is as nullable, because it defines itself as nullable implicitly. In other words, these 2 strings are equal.

Now the question whether we should or shouldn't type explicit nullable types is still open for me.

src/Config/Parser/MetadataParser/ClassesTypesMap.php
```diff
@@ -0,0 +1,76 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Overblog\GraphQLBundle\Config\Parser\MetadataParser;
+
+class ClassesTypesMap
+{
+    protected array $classesMap = [];
```
murtukov, line 9:
```suggestion
    /*
     * @var array{class: string, type: string}[]
     */
    protected array $classesMap = [];
```

If the array consists of only several keys and you know them in advance, then you can describe its exact shape ([documentation](https://phpstan.org/writing-php-code/phpdoc-types#array-shapes)). This makes the code more robust and gives more information for developers, but it's totally optional. You don't have to accept this suggestion.

src/Config/Parser/MetadataParser/MetadataParser.php
```diff
@@ -0,0 +1,956 @@
...
+            $typeConfiguration['fieldsDefaultAccess'] = self::formatExpression($accessMetadata->value);
+        }
+
+        return ['type' => $typeAnnotation->isRelay ? 'relay-mutation-payload' : 'object', 'config' => $typeConfiguration];
+    }
+
+    /**
+     * Create a GraphQL Interface type configuration from metadatas on properties.
+     */
+    private static function typeInterfaceMetadataToGQLConfiguration(ReflectionClass $reflectionClass, Metadata\TypeInterface $interfaceAnnotation): array
```
murtukov, line 367: Same here, this method always returns arrays with only 2 keys. You can describe the exact array shape and it would give PHPStan more information and make it's analysis more precision.

```suggestion
    /**
     * Create a GraphQL Interface type configuration from metadatas on properties.
     *
     * @return array{type: string, config: array}
     */
    private static function typeInterfaceMetadataToGQLConfiguration(ReflectionClass $reflectionClass, Metadata\TypeInterface $interfaceAnnotation): array
```

In this particular example you can even use a scalar type, since the value of the key 'type' is always 'interface' ([docs](https://phpstan.org/writing-php-code/phpdoc-types#literals-and-constants)):
```
/*
 * @return array{type: 'interface', config: array}
 */
```
But I don't know if scalar values can be mixed with array shapes syntax and moreover if it's necessary, because PHPStan might extract this information from the PHP's return statement directly.

src/Config/Parser/MetadataParser/MetadataParser.php
```diff
@@ -0,0 +1,956 @@
...
+        $values = [];
+
+        foreach ($reflectionClass->getConstants() as $name => $value) {
+            $valueAnnotation = current(array_filter($enumValues, fn ($enumValueAnnotation) => $enumValueAnnotation->name === $name));
+            $valueConfig = [];
+            $valueConfig['value'] = $value;
+
+            if ($valueAnnotation && isset($valueAnnotation->description)) {
+                $valueConfig['description'] = $valueAnnotation->description;
+            }
```
murtukov, line 433: Note: when I see this code 
```php
if ($valueAnnotation && isset($valueAnnotation->description))
```
it's hard for me to understand, what did you mean with $valueAnnotation. Did you want to check if it's false? Null? An empty string? A null string "0"? An integer null? An empty array?

And in order to understand it, i need to analyse the code above:
```php
$valueAnnotation = current(array_filter($enumValues, fn ($enumValueAnnotation) => $enumValueAnnotation->name === $name));
```
Okay, I see it uses `array_filter`, then it's an array, then it uses `current`, which means I need to analyse, what values can `$enumValues` contain etc. This costs other developers times, plus loose comparisons increases the possibility of errors. 

That's another reason why I encourage you to use strict comparison, whenever possible and of course until it doesn't make the code unreadable 😉

murtukov:
@Vincz we decided to release a transition version 0.14 at the same time with 1.0, since [this PR](https://github.com/overblog/GraphQLBundle/pull/786) contains many breaking changes. The 0.14 will contain deprecations only, so that the old code still works and makes the upgrade to 1.0 easier. Please check out the PR and make changes in your code accordingly, when the PR is merged.

murtukov:
As for the attributes and the '@Arg' annotation, you can use an inline syntax for function params:

Annotations:
```php
/**
 * @GQL\Field(name="friends", type="[Hero]")
 * @GQL\Arg("limit")
 * @GQL\Arg("name")
 */
public function getFriends(int $limit, string $name) {
    return array_slice($this->friends, 0, $limit);
}
```
Attributes:
```php
#[GQL\Field(name="friends", type="[Hero]")]
public function getFriends(#[GQL\Arg] int $limit, #[GQL\Arg] string $name) {
    return array_slice($this->friends, 0, $limit);
}
```

The syntax highlighting for PHP8 is still unavailable on GitHub so it might look stupid here.

Vincz:
As always, thanks for the great review @murtukov and for your time. 
I fixed a bunch of your suggestions.

About the annotations and attributes, I wanted to keep them in sync, but I have two problems:
1. The first one is that we cannot configure the default attribute with doctrine annotation. It is always `value` (see https://github.com/doctrine/annotations/issues/396). I want to have more explicit attribute names and support the short syntax (`@GQL\Field("myField")` instead of `@GQL\Field(name="MyField")`. So I have to allow the attribute to be set either by it's proper name or... by the name `value`. That's why sometimes the argument is nullable in the argument list but is not as a property. As soon as it is fixed in the doctrine annotation, I'll be able to cleanup the annotation code.
2. The second one is that doctrine annotations support nested annotations, but PHP 8 attributes don't and they support different targets. About the nested part, It's not really a problem as all annotation have been flattened so we don't use nested annotations anymore. But there is still a few issues. PHP Attributes support attributes on constants and on arguments, but doctrine annotations don't. So for your suggestion about the `@Arg`, I could target arguments, but then the attribute would not be repeatable anymore (`Attribute::IS_REPETEABLE`). It's the same problem for `@EnumValue` on constants. 

So at the end, the question is, do we keep the syntax exactly in sync between annotations and attributes to ease the upgrade process or do we embrace the full potential of attributes but with a different syntax in a few places. What do you think?

About the v0.14, I'll update the code when your PR will be merged. On my side, I have a bunch of deprecations but I still support the old syntax. I just need to support the old attributes name on `@FieldsBuilder` but the rest should still work fine.

murtukov:
@mcg-web I would definitely say annotations and attributes should go separate ways. We shouldn't try to keep them in sync only because they look similar or use similar syntax. We always try to keep all type definition syntaxes as similar as possible, but never did it to the detriment of the power of any syntax. So why would we do it now with annotations and attributes? They are completely independent and should be considered as such.

Moreover attributes are intendet to completely replace annotations, so I view them as something to be obsolete soon. Doctrine ORM has in plans to use attributes.

Now let's say a user installs the GraphQLBundle and decides to use attributes for type definitions. Why does he need the annotations code if he doesnt use it? Separating attributes and annotations would make the corresponding modules lighter.

Vincz:
> @mcg-web I would definitely say annotations and attributes should go separate ways. We shouldn't try to keep them in sync only because they look similar or use similar syntax. We always try to keep all type definition syntaxes as similar as possible, but never did it to the detriment of the power of any syntax. So why would we do it now with annotations and attributes? They are completely independent and should be considered as such.
> 
> Moreover attributes are intendet to completely replace annotations, so I view them as something to be obsolete soon. Doctrine ORM has in plans to use attributes.
> 
> Now let's say a user installs the GraphQLBundle and decides to use attributes for type definitions. Why does he need the annotations code if he doesnt use it? Separating attributes and annotations would make the corresponding modules lighter.

Makes perfect sense. It is the initial path I took but changed my mind after seeing this: https://github.com/doctrine/orm/pull/8266/files
In doctrine, they are using the same classes to handle annotations and attributes.
So, in this PR, I did the same. Annotations and attributes are sharing the same namespace, the same classes, the same attributes and the `MetadataParser`. The only thing that change between the two is the way the metadata are loaded. 
I agree that having separate bundles would be better, but there is still a lot of common code between annotations & attributes and it would be a pain to maintain two different bundles with 90% of common code.

So we would have 3 bundles:
- GraphQLAnnotationsBundle
- GraphQLAttributesBundle
- GraphQLMetadataBundle

So, to conclude, I would also vote to go separate ways because as you say, attributes are intended to replace annotations.
@mcg-web Is it fine by you?

If we all agree, I'll update the code. Maybe we could to the split in bundles later as it implies to rework the config parser to make it modular.

murtukov:
@Vincz we don't have to make metadata as a bundle. It could be a regular library just to share the common code between these 2 bundles.

As for now it totally fine how you implemented it. Lets just merge it and separate it when we start creating modules.

Vincz:
@murtukov Ok then. So, I will only update the code so annotations and attributes can already have their own syntax so I'll be able to already use the proper features of attributes and cleanup the code a bit.
We will have for example:

```
/**
 * @GQL\Enum
 * @GQL\EnumValue(name="MYVALUE", description="my description")
 */
public class MyType {
    const MYVALUE = 1;
}
```
but
```
#[GQL\Enum]
public class MyType {
  #[GQL\Description("My description")]
  const MYVALUE = 1;
}

**Comment by Vincz** [comment]
So, I removed the annotation shortcuts as it was really making the code messy. I'll wait for the proper implementation in doctrine/annotation to add the short syntax.
I removed the BC break and handle the deprecations more properly.
I didn't implement the `#[GQL\Arg]` on parameter as I'm not sure about the logic at the moment. 


@murtukov 
I also started to check how we could split the annotations and attributes, but the code is in fact almost 100% identical (as the nested annotations have been deprecated). The only difference is for the `Enum` where we need `@EnumValue` with annotation but where the attribute `#EnumValue` doesn't exist as we can use `#[GQL\Description|Deprecated]` directly on the constants. The rest is exactly the same. 

I also wanted to split the code, but honestly, after an afternoon working on it, I'm not sure anymore. Being able to use the same classes for both annotations & attributes simplify everything. If we don't, but still want to keep a shared library, we would have to add interfaces on all annotations classes and attributes classes, so it would triple the size of the code base:

Instead of just:
```
<?php

declare(strict_types=1);

namespace Overblog\GraphQLBundle\Annotation;

use Attribute;
use Doctrine\Common\Annotations\NamedArgumentConstructorAnnotation;

/**
 * Annotation for GraphQL access on fields.
 *
 * @Annotation
 * @Target({"CLASS", "PROPERTY", "METHOD"})
 */
#[Attribute(Attribute::TARGET_CLASS | Attribute::TARGET_METHOD | Attribute::TARGET_PROPERTY)]
final class Access extends Annotation implements NamedArgumentConstructorAnnotation
{
    /**
     * Field access.
     */
    public string $value;

    public function __construct(string $value)
    {
        $this->value = $value;
    }
}
```

We would need something like:

```
<?php

declare(strict_types=1);

namespace Overblog\GraphQLBundle\Annotation;

use Doctrine\Common\Annotations\NamedArgumentConstructorAnnotation;

/**
 * Annotation for GraphQL access on fields.
 *
 * @Annotation
 * @Target({"CLASS", "PROPERTY", "METHOD"})
 */
final class Access extends Annotation implements MetadataAccess, NamedArgumentConstructorAnnotation
{
    /**
     * Field access.
     */
    public string $value;

    public function __construct(string $value)
    {
        $this->value = $value;
    }

    public function getValue():string
    {
        return $this->value;
    }
}
```
and
```
<?php

declare(strict_types=1);

namespace Overblog\GraphQLBundle\Attribute;

use Attribute;

/**
 * Attribute for GraphQL access on fields.
 */
#[Attribute(Attribute::TARGET_CLASS | Attribute::TARGET_METHOD | Attribute::TARGET_PROPERTY)]
final class Access extends Attribute implements MetadataAccess
{
    public function __construct(public string $value)
    {}

    public function getValue():string
    {
        return $this->value;
    }
}
```
and
```
<?php

declare(strict_types=1);

namespace Overblog\GraphQLBundle\Shared;

/**
 * Metadata describing an field or type access
 */
interface MetadataAccess
{
    abstract public function getValue():string;
}
```

About the fact that someone using attributes would also have to download the annotations, the impact on the size is almost neglectable (just an `@Attribute` and a `@Target` or a `#[xxx]`) in the class. And as the annotations and attributes are almost identical, I'm not sure we would implement something for the attributes but not for the annotations (or otherwise). 
It still sucks to use the same namespace for both annotations & attributes, but maybe we could just add a `GraphQLMetadataBundle` with both annotations & attributes and change the namespace to `Metadata` instead of `Annotation` or `Attribute`.

**Comment by murtukov** [comment]
@Vincz well if both approaches use almost the 100% same codebase, then let it be a single bundle 🙂

One thing I don't understand is why removing shorthands? As I know the Symfony's Router uses both annotations and attributes with a single class and it's still possible to use shorthands. Why can't we do the same?

Here is the implementation: https://github.com/symfony/routing/blob/5.x/Annotation/Route.php

**Comment by Vincz** [comment]
@murtukov Yes, at the end it was exactly the same: Metadata stored on a PHP classes, but the exploitation of this metadata is the same. So there is just a very few differences than we can handle in the parser itself. 

I removed the shorthands for now, cause I didn't like the implementation. The current route annotation use the "old" way to setup the annotation. It takes the parameters from an array in the first argument of the constructor. So there is no argument validation by the annotation system. So, they introduce the `NamedArgumentConstructor` to ease the migration to attribute. But the Symfony team is also planning to use the "new" way (see https://github.com/doctrine/annotations/pull/391). 
I'll put them back as soon as we have the ability to set a default property other than `value` in annotations. It should not be a problem to add it later as there will be no BC break. I also wanted to rename a bunch of properties for more explicit names, like `expression` instead of `value` in `@Access` or `@Public` for example. 
To handle it, I was adding a `value` property to the constructor to handle the default and use it's value as the value of my default property. But I had to raise an exception if both where used and I had to mark required properties as nullable and I didn't liked it.
But the shorthand works with attributes as long as the arguments order is correct. It's just a constructor call with named arguments.

File: tests/Config/Parser/fixtures/annotations/Type/Lightsaber.php [hunk 24]
```diff
@@ -19,20 +19,23 @@ class Lightsaber
...
+    // @phpstan-ignore-next-line
     protected $color;
 
     /**
      * @ORM\Column(type="text")
      * @GQL\Field
      */
     #[GQL\Field]
+    // @phpstan-ignore-next-line
     protected $text;
```
murtukov:
```suggestion
    /**
     * @ORM\Column(type="text")
     * @GQL\Field
     * @phpstan-ignore-next-line
     */
    #[GQL\Field]
    protected $text;
```

Should also work

Vincz: I tried but it doesn't work. It seems to be because of the attribute line.

Vincz:
@murtukov About the tests, the coverage drops cause the attributes test are skipped. 
What do you think we should do about it? Should we update the coverage environment to use PHP 8?

Vincz:
Well... The tests with coverage in PHP8 is ok, but the coverage upload to scrutinizer fails.

murtukov:
@Vincz seems like switching to PHP 8 environment is the only solution

Vincz:
@mcg-web Hey man! Any idea what the coverage upload would fail? I don't want to mess with the ci tasks so I didn't try to fix it myself...

murtukov:
@Vincz I don’t know, I never configured GitHub workflows. Try setting the PHP version to “8.0” instead of just “8”

Vincz:
Ok, so the problem is ocular (the binary to upload the coverage) downloaded from the scrutinizer website is an old version (see https://github.com/scrutinizer-ci/ocular/issues/53). 
The solution seems to install it as a composer dependency. 
```
- name: "Install Ocular as depencies"
   run: composer req "scrutinizer/ocular" --dev --no-update
```
I tried, it installs fine, but I'm unable execute it, no matter the syntax I use (the binary is stored in `vendor/bin/ocular`)
I tried the following:

```
run: php vendor/bin/ocular code-coverage:upload --format=php-clover build/logs/clover.xml

run: vendor/bin/ocular code-coverage:upload --format=php-clover build/logs/clover.xml

run: ./vendor/bin/ocular code-coverage:upload --format=php-clover build/logs/clover.xml
```

But none of them worked and the binary is not found. Seems like a context problem.
ping @mcg-web

murtukov:
@Vincz the path is fixed and seems to work now. Just need to add/ignore some test coverages. 

P.S. Could you register in [Telegram](https://telegram.org/) for quick contacts?

Vincz:
@murtukov You're the best! I'll check the missing tests.
I just registered to Telegram and sent you an email with my account details.

murtukov (approved):

PR #806, opened 2021-01-11
Fix annotation for non-nullable array of non-nullable values as query argument

dmitry-hordinsky:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| RFC? | yes
| Version       | 0.13.5

When i use annotation to map argument as "Non-nullable array of non-nullable values" the scheme is built correctly. But i got error during query execution:

`GraphQL\Error\InvariantViolation: Type loader is expected to return type "Int!", but it returned ""[0]`

Example of mapping:
```php
/**
 * @GQL\Query(type="[Order]", name="order",
 *     args={
 *         @GQL\Arg(name="orderIds", type="[Int!]!"),
 *     }
 * )
 */
public function getOrders(array $orderIds): Promise
{
    ...
}
```

### DIFF
--- a/src/Transformer/ArgumentsTransformer.php
+++ b/src/Transformer/ArgumentsTransformer.php
@@ -132,7 +132,12 @@ class ArgumentsTransformer
     {
         $isRequired = '!' === $argType[strlen($argType) - 1];
         $isMultiple = '[' === $argType[0];
-        $endIndex = ($isRequired ? 1 : 0) + ($isMultiple ? 1 : 0);
+        $isStrictMultiple = false;
+        if ($isMultiple) {
+            $isStrictMultiple = '!' === $argType[strpos($argType, ']') - 1];
+        }
+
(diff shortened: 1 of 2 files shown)


Vincz:
Hey @dmitry-hordinsky! Good catch! Would you mind to add a test for this please?

dmitry-hordinsky:
@Vincz added some tests for updated fragment.

Vincz (approved):

Vincz:
Thank you @dmitry-hordinsky !

mcg-web:
@Vincz  should we also patch from `0.12` ?

Vincz:
Hey @mcg-web :-) Yes, you can. It fixes the Arguments Transformer without any BC.

mcg-web:
Thanks @Vincz for your answer, we should patch from the lower maintained version branch where the bug was introduce (0.12 in this case). We always merge branches from the older version to the newest (0.11 -> 0.12 -> 0.13 -> ... -> master). so patch will be automatically apply to other newest versions using this flow. I'll add a note about this in contributing file.

PR #818, opened 2021-02-01
Add annotation reader cache

Vincz:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | #... 
| License       | MIT

Add an cache for annotations reader to improve performances.


ferrastas:
What about allowing the user to use any implementation of `Doctrine\Common\Cache\Cache`? Even having `apcu` available on the system, the user might want to rely on the `Doctrine\Common\Cache\PhpFileCache` and take advantage of `OPcache`.

Vincz:
> What about allowing the user to use any implementation of `Doctrine\Common\Cache\Cache`? Even having `apcu` available on the system, the user might want to rely on the `Doctrine\Common\Cache\PhpFileCache` and take advantage of `OPcache`.

Hi Ferrastas, we are currently working on the next version and in the next version, the reader will be a service, so it will be easier to override it.

src/Config/Parser/AnnotationParser.php
```diff
@@ -32,18 +37,28 @@ protected static function getMetadatas(Reflector $reflector): array
...
 
-    protected static function getAnnotationReader(): AnnotationReader
+    protected static function getAnnotationReader(): Reader
     {
         if (null === self::$annotationReader) {
             if (!class_exists(AnnotationReader::class) ||
                 !class_exists(AnnotationRegistry::class)) {
                 // @codeCoverageIgnoreStart
-                throw new RuntimeException('In order to use graphql annotation, you need to require doctrine annotations');
+                throw new RuntimeException('In order to use graphql annotations, you need to require doctrine annotations');
```
murtukov, line 46:
```suggestion
                throw new RuntimeException("In order to use annotations, you need to install 'doctrine/annotations' first. See: 'https://www.doctrine-project.org/projects/annotations/1.10.html'");
```

src/Config/Parser/AnnotationParser.php
```diff
@@ -6,18 +6,23 @@
...
+use Doctrine\Common\Annotations\Reader;
+use Doctrine\Common\Cache\ApcuCache;
+use Doctrine\Common\Cache\PhpFileCache;
 use Overblog\GraphQLBundle\Config\Parser\MetadataParser\MetadataParser;
 use ReflectionClass;
 use ReflectionMethod;
 use ReflectionProperty;
 use Reflector;
 use RuntimeException;
+use function apcu_enabled;
```
murtukov, line 19:
```suggestion
use function apcu_enabled;
use function md5;
```

murtukov:
@Vincz I pushed some changes into your branch. Don't know what is the problem with Ocular. Didn't you already have this problem with Ocular?

Vincz:
> @Vincz I pushed some changes into your branch. Don't know what is the problem with Ocular. Didn't you already have this problem with Ocular?
Hi @murtukov  :-) 
I had a problem with it before, yes. I think, it was because it was not running from the main repository if I remember correctly. But curiously, others PR seem to work just fine. 
Seems to be a GIT problem, but my skills with it are pretty limited...

murtukov:
@Vincz do you remember which PR was it, where we fixed the issue with Ocular?

Vincz:
Handle by #867

Issue #843, opened 2021-05-07
Simplify TypeGenerator dependencies


murtukov:
| Q                | A
| ---------------- | -----
| RFC?             | yes
| Version/Branch   | 0.14

The TypeGenerator class has a lot of dependencies and the most of them are Symfony parameters. The current service declaration lloks like this:
https://github.com/overblog/GraphQLBundle/blob/f0af7475e829b0ed196732fae84b90ed58fe68b9/src/Resources/config/services.yaml#L63-L72

and the class constructor with properties looks like this:
https://github.com/overblog/GraphQLBundle/blob/f0af7475e829b0ed196732fae84b90ed58fe68b9/src/Generator/TypeGenerator.php#L30-L64

in my opinion it looks messy. What I would like to do is to pass all the parameter dependencies together, either as an array or an object and maybe rename some of them to make them more descriptive. I considered 3 different ways to do that, but I will describe here only 1 of them, which I think is the best way:

Change the service declaration to pass all the parameters as an array:
```yaml
Overblog\GraphQLBundle\Generator\TypeGenerator:
    arguments:
        - '@Overblog\GraphQLBundle\Generator\TypeBuilder'
        - '@Symfony\Contracts\EventDispatcher\EventDispatcherInterface'
        - classNamespace: "%overblog_graphql.class_namespace%"
          cacheDir:       "%overblog_graphql.cache_dir%"
          config:         "%overblog_graphql_types.config%"
          useClassMap:    "%overblog_graphql.use_classloader_listener%"
          baseCacheDir:   "%kernel.cache_dir%"
          cacheDirMask:   "%overblog_graphql.cache_dir_permissions%"
```

class constructor:
```php
public function __construct(TypeBuilder $typeBuilder, EventDispatcherInterface $eventDispatcher, array $options) 
{
    $this->typeBuilder = $typeBuilder;
    $this->eventDispatcher = $eventDispatcher;
    $this->options = new GeneratorOptions($options);
}
```
As you can see, the class has now only 3 dependencies. Parameters are combined into the `$options` array and passed into `GeneratorOptions` class which is just a DTO to avoid working with arrays. It destructures the array and creates a nice object with autocompletion. Also since properties are typed, there is no need to create getters/setters:

```php
class GeneratorOptions
{
    public ?string $cacheDir;
    public int $cacheDirMask;
    public array $configs;
    public bool $useClassMap;
    public ?string $baseCacheDir;
    public string $classNamespace;

[...]

PR #845, opened 2021-05-09
Refactor `TypeGenerator`

murtukov:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | #843 

This is a draft.

@mcg-web I think the GraphQL configs (`$types` argument) should be moved away from `TypeGeneratorOptions` because it's not an option, but actual data the generator works with.

I also kept the constructor in the `TypeGeneratorOption` cause it ensures the object to be properly initialized. Without a constructor the object can have uninitialized properties, which is not desired:

```php
$options = new TypeGeneratorOptions();
$options->cacheDir = "...";

$myVal = $options->cacheDirMask; // throws an error, property uninitialized
```

The `getCacheDir` method was simplified and renamed into `getCacheDirOrDefault`. The argument is not needed anymore, because if you wanna access the `cacheDir` without the fallback to default you can simply access it directly: `$options->cacheDir`


src/Generator/TypeGeneratorOptions.php
```diff
@@ -0,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Overblog\GraphQLBundle\Generator;
+
+class TypeGeneratorOptions
+{
+    public string $namespace;
```
mcg-web, line 9: public properties can be error prone here (same reason as above).

murtukov:
@mcg-web ok, so now the options cannot be accessed outside of `TypeGenerator` class, except the ones that are exposed via methods: 

- `TypeGenerator::getCacheDir()`
- `TypeGenerator::setCacheDirMask()`
- `TypeGenerator::setCacheBaseDir()`
- `TypeGenerator::getCacheBaseDir()`

---

What about this?

> I think the GraphQL configs ($types argument) should be moved away from TypeGeneratorOptions because it's not an option, but actual data the generator works with.

mcg-web:
>I think the GraphQL configs ($types argument) should be moved away from TypeGeneratorOptions because it's not an option, but actual data the generator works with.

We can indeed inject $types directly in TypeGenerator.

mcg-web (approved):

Issue #853, opened 2021-05-24
Paginator class depends on internal state


piotrkreft:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | yes
| BC Break report? | no
| RFC?             | no
| Version/Branch   | 0.13

Class `Overblog\GraphQLBundle\Relay\Connection\Paginator` depends on internal totalCount state, if fetched already. I find this to be sort of an antipattern in several situations, especially when dealing with DI and class being a service as it brings an additional potential place of failure being harder to find in simple tests even.

Would you consider a change? If yes I'd obviously provide a suitable code for it.
If BC desired the solution could be configurable behavior via an optional constructor parameter though not sure whether it's a good practice tbh. Alternatively, an interface and cache decorator would be more suitable perhaps.

Best regards

piotrkreft:
As a suggested solution I decided to go with a fully compatible approach and ready out-of-the-box cache callable.
Callable implements ResetInterface so it does grant the developer a possibility to reset cache state and also will be reused by Symfony HTTPKernel if used as service to be reset per request - desired if for an instance one would user DI service and long running process like Roadrunner etc.

PR #855, opened 2021-05-29
Total count cache

piotrkreft:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | #853
| License       | MIT

The goal of the proposal is to give the possibility to decide about the caching strategy for the developers which in some cases may be unwanted

### DIFF
--- a/docs/helpers/relay-paginator.md
+++ b/docs/helpers/relay-paginator.md
@@ -244,6 +244,26 @@ class Greetings implements QueryInterface
 }
 ````
 
+#### Total count caching
+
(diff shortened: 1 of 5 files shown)


piotrkreft:
Unit tests issues causing build to fail are unrelated to the change and fail on the `master` branch as well.
Should I give it a go and try to fix'em or this should be done as a separate PR?

PR #864, opened 2021-06-17
Fix PHP 8.0 deprecation in uksort

### LINKED ISSUE #831: Deprecation notice on PHP 8
Hi, the following deprecation warning is triggered on PHP 8:

> Deprecated: uksort(): Returning bool from comparison function is deprecated, return an integer less than, equal to, or greater than zero in /app/vendor/overblog/graphql-bundle/src/Config/Processor/InheritanceProcessor.php on line 89

Using `v0.13.5`.

gnoddep:
Since PHP 8.0 it is deprecated behaviour for the callback in `uksort` to
return a boolean and it must return an integer.

Fixes #831

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | #831
| License       | MIT

### DIFF
--- a/src/Config/Processor/InheritanceProcessor.php
+++ b/src/Config/Processor/InheritanceProcessor.php
@@ -85,7 +85,7 @@ final class InheritanceProcessor implements ProcessorInterface
 
         // Restore initial order
         \uksort($parentTypes, function ($a, $b) use ($parents) {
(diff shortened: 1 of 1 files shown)


gnoddep:
Fixed with #865

PR #865, opened 2021-06-21
Fix uksort() callback returning bool deprecation

BenMorel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Documented?   | no
| Fixed tickets | #831
| License       | MIT

Fixes a deprecation on PHP 8.0:

> uksort(): Returning bool from comparison function is deprecated, return an integer less than, equal to, or greater than zero

This targets `0.12` as the `uksort()` call was introduced in this branch.

Not sure why all tests fail: failures are unrelated to this commit (I tried reverting it, same result).


mcg-web (approved):

mcg-web:
Thank you!

PR #872, opened 2021-07-22
fix(ErrorHandler): Argument 3 passed to Overblog\GraphQLBundle\Error\ErrorHandler::__construct() must be of the type string, null given

Oliboy50:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes/no
| Documented?   | yes
| Fixed tickets | #... 
| License       | MIT

because, I have the following error during `KernelTerminate` event:

> Argument 3 passed to Overblog\GraphQLBundle\Error\ErrorHandler::__construct() must be of the type string, null given

using the following YAML bundle configuration:
```yaml
overblog_graphql:
    # ...
    errors_handler:
        enabled: true
        internal_error_message: ~
        rethrow_internal_exceptions: true
        debug: true
        log: false
        logger_service: logger
 ```

using Symfony `v4.4.26` and GraphQLBundle `v0.13.6`

### DIFF
--- a/src/DependencyInjection/OverblogGraphQLExtension.php
+++ b/src/DependencyInjection/OverblogGraphQLExtension.php
@@ -196,10 +196,12 @@ class OverblogGraphQLExtension extends Extension
             ->setArgument(0, $this->buildExceptionMap($config['errors_handler']['exceptions']))
             ->setArgument(1, $config['errors_handler']['map_exceptions_to_parent']);
 
-        $container->register(ErrorHandler::class)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by Oliboy50** [comment]
closing because I think that the correct fix would be to simply remove the `internal_error_message` key from the bundle configuration (maybe the documentation should be updated accordingly)

because:
```php
->scalarNode('internal_error_message')->defaultValue(ErrorHandler::DEFAULT_ERROR_MESSAGE)->end()
```

PR #880, opened 2021-09-22
0.12

Vincz:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| Tests pass?   | yes

Fix doctrine column type guessing when no annotation is set (default to string).


mcg-web (approved):
Thank you @Vincz

PR #888, opened 2021-10-11
Remove `addDefaultFallBackToTypeLoader`

### LINKED ISSUE #886: Remove `ExtensibleSchema::addDefaultFallBackToTypeLoader`
While working on performance improvements for our large GraphQL application I noticed something odd. 

I disabled `config_validation`, `use_classloader_listener` and `auto_compile` and noticed that for some queries `\GraphQL\Type\Schema::getTypeMap` was called. The description of that method warns that it's slow: "This operation requires full schema scan. Do not use in production environment.".

I notice this was because of a misconfiguration in my application. It couldn't find an EnumType and would then fallback to calling `getType` that would eventually call `getTypeMap`. That was able to find the type and register it properly.

See 
https://github.com/overblog/GraphQLBundle/blob/886cd9d29beb8aeb537490b1d3e63a33327a7331/src/Definition/Type/ExtensibleSchema.php#L66-L74

Why is this fallback in place? Wouldn't it be better to fail immediately so that developers know their configuration is not right. Having these things automagically resolved is nice, but does not benefit performance and leaves bad config as is.

ruudk:
Fixes #886

| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  |no
| BC breaks?    | no
| Deprecations? |no
| Tests pass?   | yes
| Documented?   |no
| Fixed tickets | #886
| License       | MIT

### DIFF
--- a/src/Definition/Type/ExtensibleSchema.php
+++ b/src/Definition/Type/ExtensibleSchema.php
@@ -4,19 +4,17 @@ declare(strict_types=1);
 
 namespace Overblog\GraphQLBundle\Definition\Type;
 
-use GraphQL\Type\Definition\Type;
 use GraphQL\Type\Schema;
 use GraphQL\Type\SchemaConfig;
 use Overblog\GraphQLBundle\Definition\Type\SchemaExtension\SchemaExtensionInterface;
-use Overblog\GraphQLBundle\Resolver\UnresolvableException;
 
 class ExtensibleSchema extends Schema
 {
     public function __construct($config)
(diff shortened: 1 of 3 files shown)


ruudk:
I changed the target to `master` and there is only one test that fails: `MultipleSchemaTest`. 

@mcg-web Any idea how to solve this?

mcg-web (approved):

mcg-web:
Thank you @ruudk !

PR #890, opened 2021-10-13
Always compile types with `AliasedInterface`

ruudk:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | ?
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | 
| License       | MIT

I noticed that types generated by `graphql:compile` don't include the `AliasedInterface`. This is weird, because in my YAML definition I reference everything with aliases.


ruudk:
@murtukov Applied your feedback.

PR #905, opened 2021-11-18
Fix DebugCommand dependency injection

jdecool:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Documented?   | no
| Fixed tickets | N/A
| License       | MIT

In one of our projects, we decorate `QueryResolver` and `MutationResolver` for our needs.

But currently, the `DebugCommand` depends on concrete implementations instead of interfaces.
So when we run the `bin/console list`, Symfony crash because of the dependency injection.

This PR updates the `DebugCommand` constructor to inject `FluentResolverInterface` instead of concrete implementations.

### DIFF
--- a/src/Command/DebugCommand.php
+++ b/src/Command/DebugCommand.php
@@ -5,9 +5,6 @@ declare(strict_types=1);
 namespace Overblog\GraphQLBundle\Command;
 
 use Overblog\GraphQLBundle\Resolver\FluentResolverInterface;
-use Overblog\GraphQLBundle\Resolver\MutationResolver;
-use Overblog\GraphQLBundle\Resolver\ResolverResolver;
-use Overblog\GraphQLBundle\Resolver\TypeResolver;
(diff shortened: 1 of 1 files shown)


ruudk (approved):

mcg-web:
This change can introduce a break change before php 8, can we introduced this from 0.14 @jdecool ?

jdecool:
@mcg-web It could.

But I don't understand why it could introduce a BC before PHP 8. 
All class used implement the `FluentResolverInterface` so they can be replaced by the interface they use no ?

ruudk:
The class it not final, so it could be extended. But it shouldn't be an issue, see https://3v4l.org/OZ86V

mcg-web:
My bad since it's the construct , it is not consider as a BC.

PR #933, opened 2021-12-02
Use Flex to require specific Symfony versions

ruudk:
This is the recommended way to pin everything to the same major.minor. 

Without it, it will still install other packages.


ruudk:
@mcg-web Ready for review :) By using Flex we now make sure that we pin all Symfony components to the same major.minor version. Before, it wouldn't work as we thought.

mcg-web (approved):

mcg-web:
Thank you @ruudk !

PR #937, opened 2021-12-05
Add CI matrix Webonyx graphql-php versions

mcg-web:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | none
| License       | MIT


.github/workflows/ci.yml
```diff
@@ -23,6 +23,14 @@ jobs:
...
           - '5.3'
           - '5.4'
+        graphql-version:
+          - '14.5.*'
+          - '14.6.*'
+          - '14.7.*'
+          - '14.8.*'
+          - '14.9.*'
+          - '14.10.*'
+          - '14.11.*'
```
ruudk, line 33: Why test all versions? Wouldn't lowest and highest suffice?

ruudk: Because the matrix will be huge this way, if you take into account Symfony 4.4, 5.3, 5.4 and soon 6.0 + php versions (7.4, 8.0 and soon 8.1).

mcg-web: Yes lowest and highest should suffice indeed.

PR #941, opened 2021-12-06
Replace deprecated `NamedArgumentConstructorAnnotation` with `@NamedArgumentConstructor` + Address one PHP 8.1 deprecation

ruudk:
Also fixed a problem that will pop-up in PHP 8.1 with the `TypeInterface`:

In PHP 8.1 optional parameters cannot be defined before required parameters.

Because `string name = null` is optional, we also have to make `string $resolveType` optional. This way it works on PHP 8.1 without breaking BC.

In 1.0, we should change the order.

### DIFF
--- a/src/Annotation/Access.php
+++ b/src/Annotation/Access.php
@@ -5,16 +5,17 @@ declare(strict_types=1);
 namespace Overblog\GraphQLBundle\Annotation;
 
 use Attribute;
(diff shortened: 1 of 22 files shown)


ruudk:
@mcg-web Please have a look.

If you merge this, please also update master so that I can immediately address the deprecation there.

mcg-web:
done @ruudk !

PR #969, opened 2021-12-18
Added async support to the interface.

benelori:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | #... 
| License       | MIT

This PR adds support for ReactPHP promises, because the wait() method of the current executor is synchronous and it will block the event loop.

This is more like a hack btw, because it's the first time I dove into the project and I need something to work for my current work, which has ReactPHP.

I have issues with adding yet another flag to be honest and I'm not sure how to solve the issue with the controller, which will need to something like this in the createResponse method

`
if ($payload instanceof Promise) {
                return new ReactPromise(
                    function ($resolve) use ($payload) {
                        $payload
                            ->then(
                                function (ExecutionResult $test) use ($resolve) {
                                    $resolve(new JsonResponse($test->toArray(), 200));
                                },
                                function ($error) {
                                    var_dump($error);
                                }
                            );
                    }
                );
            }
`
I would be glad to implement this properly, so I would like to ask for some opinions, on how this actually should look like

### DIFF
--- a/src/Controller/GraphController.php
+++ b/src/Controller/GraphController.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Overblog\GraphQLBundle\Controller;
 
+use GraphQL\Executor\Promise\Promise;
 use Overblog\GraphQLBundle\Request\BatchParser;
 use Overblog\GraphQLBundle\Request\Executor;
 use Overblog\GraphQLBundle\Request\Parser;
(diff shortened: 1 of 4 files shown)


AlexOstrovsky:
@mcg-web is there any plan to work on this? Or is this kind of functionality available?

PR #971, opened 2021-12-23
Support defining enum values by callable and/or generator

ruudk:
This PR adds support for the new feature merged in webonyx/graphql-php#1040.

It allows returning enum values by closure and/or generator.

While working on this I also noticed that the object decorator doesn't work properly when the objects fields are returned by a generator. That's now fixed as well.

I had to add a PHPStan ignore line because of this issue: https://github.com/webonyx/graphql-php/issues/1041


mcg-web (approved):

mcg-web:
Thank you @ruudk !

PR #983, opened 2022-02-02
Define default value for `$processors`

ruudk:
This makes it possible to initialize a `ConfigProcessor` that has no processors.

Solves this error:

Typed property Overblog\GraphQLBundle\Definition\ConfigProcessor::$processors must not be accessed before initialization.


mcg-web (approved):

mcg-web:
Thank you @ruudk !

Issue #991, opened 2022-03-04
Input validation generates wrong Composite validators


bravik:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | yes
| RFC?             | no
| Version/Branch   | 0.14.2

Validator definition:
```yaml
            validation:
              - AtLeastOneOf:
                  constraints:
                    - Expression: "args['from'] === null && args['to'] !== null"   # Skip if values are null
                    - GreaterThanOrEqual: '-3 months'
                  message: "Specified period is too large. Max 3 month allowed"
                  includeInternalMessages: false
```

In generated type definitions input validation for Symfony `Composite` constraints is now generated in following manner:
```php
  'args' => [
      [
          'name' => 'from',
          'type' => fn() => $services->getType('DateTime'),
          'validation' => fn() => [
              new SymfonyConstraints\AtLeastOneOf([
                  'constraints' => [[
                      'Expression' => 'args[\'from\'] === null && args[\'to\'] !== null',
                  ], [
                      'GreaterThanOrEqual' => '-3 months',
                  ]],
                  'message' => 'Specified period is too large. Max 3 month allowed',
                  'includeInternalMessages' => false,
              ]),
          ],
      ],
```
Which failes with `Array to string conversion`
Which in turn happens when `Composite` attempts to throw it's error:
```
 2 => array:3 [
        "file" => "/app/vendor/symfony/validator/Constraints/Composite.php"
        "line" => 74
        "function" => "sprintf('The value "%s" is not an instance of Constraint in constraint "%s".', array(1), 'Symfony\Component\Validator\Constraints\AtLeastOneOf')"    
      ]
```

In type definition it should be:

```
                    'args' => [
                        [
                            'name' => 'from',
                            'type' => fn() => $services->getType('DateTime'),
                            'validation' => fn() => [
                                new SymfonyConstraints\AtLeastOneOf([
                                     'constraints' => [
                                                    new SymfonyConstraints\Expression(
                                                        'args[\'from\'] === null && args[\'to\'] !== null'
                                                    ),
                                                    new SymfonyConstraints\GreaterThanOrEqual('-3 months')
                                                ],
                                     'message' => 'Specified period is too large. Max 3 month allowed',
                                     'includeInternalMessages' => false,
                                     ]),
                          ],
             ],
```

PR #992, opened 2022-03-04
Fix type generation for Composite symfony validator with explicit 'constraints' key

bravik:
Fix for https://github.com/overblog/GraphQLBundle/issues/991

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | https://github.com/overblog/GraphQLBundle/issues/991
| License       | MIT

### DIFF
--- a/src/Generator/TypeBuilder.php
+++ b/src/Generator/TypeBuilder.php
@@ -637,6 +637,9 @@ class TypeBuilder
                 if (isset($args[0]) && is_array($args[0])) {
                     // Nested instance
                     $instance->addArgument($this->buildConstraints($args, false));
+                } elseif (isset($args['constraints'][0]) && is_array($args['constraints'][0])) {
+                    // Nested instance with explicit "constraints" key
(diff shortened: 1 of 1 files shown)


mcg-web:
Thank you @bravik for the fix ! Can you please add a test covering this case?

bravik:
@mcg-web There are only 3 tests for TypeGenerator at all.. Please, I just don't have enough mental strength to do this, while all this war is going on out here...

bravik:
Updated fix to take into account other options passed to validation constraint

mcg-web:
Ok I can understand ! I will add test later with your example in #991. Thanks again for the fix and the feedback !

mcg-web (approved):

bravik:
@mcg-web will a patch version be released?

mcg-web:
released in [0.14.3](https://github.com/overblog/GraphQLBundle/releases/tag/v0.14.3)

PR #1071, opened 2022-12-02
Append @deprecated directive for Input's fields

TaProhm:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | #...
| License       | MIT

The @deprecated directive is supported by this bundle for different GraphQL entities: Types, Objects, Interfaces, Enums.
But not for Inputs for some reason.

### DIFF
--- a/src/Config/InputObjectTypeDefinition.php
+++ b/src/Config/InputObjectTypeDefinition.php
@@ -32,6 +32,7 @@ final class InputObjectTypeDefinition extends TypeDefinition
                         ->append($this->descriptionSection())
                         ->append($this->defaultValueSection())
                         ->append($this->validationSection(self::VALIDATION_LEVEL_PROPERTY))
+                        ->append($this->deprecationReasonSection())
                     ->end()
                     ->isRequired()
                     ->requiresAtLeastOneElement()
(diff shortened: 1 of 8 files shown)


TaProhm:
Hi @mcg-web and @Vincz. Could this PR be reviewed, please. The change is relatively small, but rather important for everyday work.

Vincz:
Hi @TaProhm, sorry for the late reply.
Would you mind to update this line : https://github.com/overblog/GraphQLBundle/blob/f33382ac2bbdb750cbe448ac74cb1b6207e73c4f/src/Config/Parser/MetadataParser/MetadataParser.php#L401 to pass `true` as the second argument of `getDescriptionConfiguration`.

Can you also add tests? (just to ensure that it works when we add the directive in yaml or with annotations/attributes)

TaProhm:
@Vincz Thank you for pointing out in the right direction. As a result I stuck with" Overblog\GraphQLBundle\Tests\Functional\Type\DefinitionTest" for InputObjectType. It turned out there is an issue with [webonyx/graphql-php](https://github.com/webonyx/graphql-php/) regarding input values deprecation https://github.com/webonyx/graphql-php/issues/110
I will ammend this PR with tests as soon as it will be possible.

TaProhm:
Hi [Vincz](https://github.com/Vincz)

It's only now that I've been able to give a detailed answer. I contributed to webonyx/graphql-php/issues/110 during my vacation, and then I was going to finish this PR. But now there is not enough time. However, this functionality is rather necessary, so I plan to finish it as soon as I can.

Vincz (approved):

Vincz:
Awesome @TaProhm! Thank you very much for this :)

PR #1073, opened 2022-12-13
Fix tests & arguments transformer when data is not set

Vincz:
When data for an input doesn't exist (ie. The property is not set in the data), use the default value of the input property or null if is doesn't have one.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| License       | MIT

When a input object is used in conjunction with the `ArgumentsTransformer`, if the data doesn't contain the property, use the default one from the input object instead of null.

For example:

```php
<?php
final class MyInput
{
    public $field1;

    public $field2 = "default_me";
}
?>
```
If the submitted `data` doesn't include the `field2` property, the input instance will be left untouched and `$field2` will equals `default_me` instead of `null`.

### DIFF
--- a/src/Transformer/ArgumentsTransformer.php
+++ b/src/Transformer/ArgumentsTransformer.php
@@ -102,7 +102,7 @@ final class ArgumentsTransformer
             $fields = $type->getFields();
 
             foreach ($fields as $name => $field) {
-                $fieldData = $this->accessor->getValue($data, sprintf('[%s]', $name));
+                $fieldData = \array_key_exists($name, $data) ? $this->accessor->getValue($data, sprintf('[%s]', $name)) : $this->accessor->getValue($instance, $name);
                 $fieldType = $field->getType();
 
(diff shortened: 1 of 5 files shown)


Vincz:
ping @mcg-web

mcg-web (approved):
Thank you @Vincz !

deeky666:
@Vincz @mcg-web this breaks typed Properties where no default value is set. Just encountered this while upgrading to v1.0:

```php
<?php

namespace App\Input;

use Overblog\GraphQLBundle\Annotation as GQL;

#[GQL\Input]
final class MyInput
{
    #[GQL\Field]
    public ?string $field;
}
```

Throws Error: `Typed property App\\Input\\MyInput::$field must not be accessed before initialization`

See also: https://3v4l.org/9KReD

Issue #1030, opened 2022-07-05
Missing attribute definition in EnumValue


pelex:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no
| Version/Branch   | 0.15.1

I have  `User Deprecated: The attributes "values" on annotation @GQL\Enum is deprecated as of 0.14 and will be removed in 1.0. Use the @GQL\EnumValue annotation on the class itself instead.` warning but can't use EnumValue as attribute. Missing attribute definition:
```
#[Attribute(Attribute::TARGET_CLASS | Attribute::IS_REPEATABLE)]
final class EnumValue extends Annotation
```

Vincz:
Hi @pelex! It is not possible to attach annotations to constants. That's why `@GQL\EnumValue` is needed to define additional attributes (`description` or `deprecation`) on constants of the enum class.

With attributes, we can use `#[GQL\Description]` or `#[GQL\Deprecated]` directly on the constants, so the attribute `#[GQL\EnumValue]` is not needed.

To better understand, 

Check this file https://github.com/overblog/GraphQLBundle/blob/master/tests/Config/Parser/fixtures/annotations/Enum/Race.php

And this test 
https://github.com/overblog/GraphQLBundle/blob/405e83b60af3e5fdaf909473cd05cbc53fb34d03/tests/Config/Parser/MetadataParserTest.php#L240

pelex:
Hello @Vincz , thank you for the answer. Example from this file allowed me to set all attributes correctly.

PR #1111, opened 2023-03-17
Add Attribute definition to EnumValue so it can be used as an attribute.

IzabelaTimofte:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | #1030
| License       | MIT

Added Attribute definition on EnumValue

### DIFF
--- a/src/Annotation/EnumValue.php
+++ b/src/Annotation/EnumValue.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Overblog\GraphQLBundle\Annotation;
 
+use Attribute;
 use Doctrine\Common\Annotations\Annotation\NamedArgumentConstructor;
 
 /**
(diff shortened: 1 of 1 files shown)


Vincz:
Hi @IzabelaTimofte!
Thank you for the PR. Did you checked https://github.com/overblog/GraphQLBundle/blob/master/docs/definitions/type-system/enum.md#working-with-native-php-81-enums

There is no attribute with `EnumValue` cause it was used only to attach description or deprecated reason to a value cause before PHP 8.1 and the attributes, the annotations library was not able to handle annotation on contants. That's why this `EnumValue` was used.
Now with attributes, you can attach directly the `#GQL\Description` to the enum value as in the example here: 

```php
#[GQL\Enum]
enum Color 
{
    #[GQL\Description("The color red")]
    case RED;
    case GREEN;
    case BLUE;
}
```

So attribute is useless on `EnumValue`

PR #1113, opened 2023-03-26
Ensure registerLoader method exists on AnnotationReader

Vincz:
`registerLoader` method doesn't exist anymore. We must still check for older versions.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| Deprecations? | yes
| Tests pass?   | yes/no


mcg-web (approved):
Thank you @Vincz !

PR #1115, opened 2023-04-11
fix/profiler-limit-type

clementvtrd:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Documented?   | no
| License       | MIT

I'm getting a Symfony error when opening the GraphQL tab of the Profiler. I found out the limit is now hard typed in `int` and a string is actually sent from the ProfilerController.

## Screenshot

Capture d’écran 2023-04-11 à 14 50 25

PR #1118, opened 2023-04-25
Account for type change in Symfony Profiler ^6.2

darrylhein:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | n/a
| Fixed tickets | none atm
| License       | MIT

In Symfony 6.2, the `$limit` argument to `Profiler::find()` was changed from a string to an integer in this commit: https://github.com/symfony/http-kernel/commit/13b9831be953c052c6c515a7fd58c7d31979542d and related issue: https://github.com/symfony/symfony/issues/49656 This takes into account older Symfony versions that only accept a string.


darrylhein:
Dup of #1115

Issue #721, opened 2020-08-07
Memory issue in v0.13


Kocal:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | yes/no
| RFC?             | yes/no
| Version/Branch   | 0.13

Hi,

I'm still working on GraphQLBundle update from 0.11 to 0.13 on our big app at work, but I detected a memory issue while running our Behat suite tests.

It does not affect 0.11 and 0.12.

I've recorded `htop` output while running behat:
- 0.11, the VM used memory stays under 600 MB: https://streamable.com/sjrgq0
- 0.13, the memory increase quickly to ~2 GB, and makes PHP crash: https://streamable.com/qmcnfx

Do you have an idea of what can happens?
Thanks!

mcg-web:
Hi,

`InputValidator` is the only big change between 0.12 and 0.13 but if you only updating you are not concern by this. Have you updated others dependencies (Symfony, Webonyx, ...) ?

Kocal:
While updating from 0.11.11 to 0.13.3, the following dependencies has been updated:
- `symfony/config`, `symfony/dependency-injection`, `symfony/event-dispatcher`, `symfony/filesystem`, `symfony/inflector`, `symfony/options-resolver`, and `symfony/property-access` from 4.4.10 to 4.4.11
- `symfony/event-dispatcher-contracts` from 1.1.7 to 1.1.9
- `symfony/polyfill-ctype` from 1.17.1 to 1.18.1
- `symfony/service-contracts` from 2.1.2 to 2.1.3
- `webonyx/graphql-php` from 0.12.6 to 0.13.9

Then I downgraded to overblog/graphql-bundle from 0.13.3 to 0.12.8 but no other dependencies has been downgraded... 

So it means that the issue comes from the bundle :confused:

Kocal:
I've done some tests today, and this is what I found.

The memory leak also happens when running Behat tests not on the GraphQL API. For example some _classical_ tests on our controllers (`ArticleController`):
- 0.12.8: 
  - `features/api`: 1.10GB max
  - `features/articles`: 1GB max
- 0.13.3: 
  - `features/api`: increase to 2GB, then I stop the tests
  - `features/articles`: increase to 1.48 GB, then tests stop themselves (all successful)

[...]

PR #1119, opened 2023-05-02
fix: Fix the memory leak

theofidry:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | could not check
| Documented?   | no
| Fixed tickets | #721
| License       | MIT

Follow up of [my
comment](https://github.com/overblog/GraphQLBundle/issues/721#issuecomment-1414158275).

I cannot investigate further, but my guess is that exceptions are not cheap objects. When creating an exception you also capture the context and stack trace which means that even if the extension is unsued:

- It slows down the program because instantiating it is not cheap.
- May prevent some objects to be garbage collected because referenced in the exception.

### DIFF
--- a/src/ExpressionLanguage/ExpressionFunction.php
+++ b/src/ExpressionLanguage/ExpressionFunction.php
@@ -12,7 +12,9 @@ class ExpressionFunction extends BaseExpressionFunction
     public function __construct(string $name, callable $compiler, ?callable $evaluator = null)
     {
         if (null === $evaluator) {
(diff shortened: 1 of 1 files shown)


Vincz:
Hi @theofidry!
Can you check the tests please?
Don't have time to dig into this atm but it seems to break the tests suite: https://github.com/overblog/GraphQLBundle/pull/1119/checks

Thanks!

theofidry:
Sorry will try to come back to this in a near future

PR #1151, opened 2024-01-12
Merge guessed argument properties and args described with GQL\Arg

mathroc:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | yes/no
| Deprecations? | no
| Tests pass?   | yes/no
| Documented?   | no
| Fixed tickets | #... 
| License       | MIT

When a single argument is described with `#[GQL\Arg]` all other are not guessed anymore

This commit starts by guessing all arguments and override the ones that have `#[GQL\Arg]` attributes describing them

I'm not sure of a few things:

* should the type specified in the attribute override the one that was guessed?
* same for the default
* Is there cases where guessing could fail and adding the #[GQL\Arg] attribute was a way to prevent guessing failing / crashing?

in the future it would be nice if we could add the attribute on the argument directly instead of the method so we could make the name & type optional there (or forbidden?)

### DIFF
--- a/src/Config/Parser/MetadataParser/MetadataParser.php
+++ b/src/Config/Parser/MetadataParser/MetadataParser.php
@@ -568,7 +568,7 @@ abstract class MetadataParser implements PreParserInterface
             }
 
             if (isset($arg->default)) {
-                $args[$arg->name]['defaultValue'] ??= $arg->default;
+                $args[$arg->name]['defaultValue'] = $arg->default;
             }
         }
 
--- a/src/Config/Parser/MetadataParser/MetadataParser.php
+++ b/src/Config/Parser/MetadataParser/MetadataParser.php
@@ -556,10 +556,6 @@ abstract class MetadataParser implements PreParserInterface
         /** @var Metadata\Arg[] $argAnnotations */
         $argAnnotations = self::getMetadataMatching($metadatas, Metadata\Arg::class);
 
-        if ($reflector instanceof ReflectionMethod) {
-            $args = self::guessArgs($reflectionClass, $reflector);
-        }
-
         foreach ($argAnnotations as $arg) {
             $args[$arg->name] = ['type' => $arg->type];
 
@@ -570,6 +566,10 @@ abstract class MetadataParser implements PreParserInterface
             if (isset($arg->default)) {
                 $args[$arg->name]['defaultValue'] = $arg->default;
             }
+        }
+
+        if ($reflector instanceof ReflectionMethod) {
+            $args = self::guessArgs($reflectionClass, $reflector, $args);
         }
 
         if (!empty($args)) {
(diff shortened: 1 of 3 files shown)


Vincz:
@mathroc to answer your questions:
- Should the type specified in the attribute override the guessed one => YES. The developper should have the final word regarding the argument type. So, `#[GQL\Arg]` should be the priority.
- For the default as well.
- We auto-guess only if we don't have a `#[GQL\Arg]`

You are right, we could also explore the possibility to set attributes directly on the arguments.

src/Config/Parser/MetadataParser/MetadataParser.php
```diff
@@ -556,22 +556,22 @@ private static function getTypeFieldConfigurationFromReflector(ReflectionClass $
         /** @var Metadata\Arg[] $argAnnotations */
         $argAnnotations = self::getMetadataMatching($metadatas, Metadata\Arg::class);
 
+        if ($reflector instanceof ReflectionMethod) {
+            $args = self::guessArgs($reflectionClass, $reflector);
+        }
+
         foreach ($argAnnotations as $arg) {
-            $args[$arg->name] = ['type' => $arg->type];
+            $args[$arg->name] ??= ['type' => $arg->type];
```
mathroc, line 564:
```suggestion
            $args[$arg->name] = ['type' => $arg->type];
```

src/Config/Parser/MetadataParser/MetadataParser.php
```diff
@@ -556,22 +556,22 @@ private static function getTypeFieldConfigurationFromReflector(ReflectionClass $
...
 
             if (isset($arg->description)) {
                 $args[$arg->name]['description'] = $arg->description;
             }
 
             if (isset($arg->default)) {
-                $args[$arg->name]['defaultValue'] = $arg->default;
+                $args[$arg->name]['defaultValue'] ??= $arg->default;
```
mathroc, line 571:
```suggestion
                $args[$arg->name]['defaultValue'] = $arg->default;
```

mathroc (commented):
make the attributes have the final word on type and default value

mathroc:
and for the last question: yes sometimes the guessing can't work at all, e.g.: arrays

:thinking:

Vincz:
Yes, and we want an error when we don't have a corresponding `#[GQL\Arg]` and the auto-guess failed.  
To make this work, you first need to get arguments having a corresponding`#[GQL\Arg]` and auto-guess only the remaining ones (by passing the list of already resolved to the auto-guesser for example).

Vincz:
@mathroc The GraphQL arguments order doesn't matter as they are named. But we need them in the right order when the PHP method is called.

Vincz (approved):

Issue #1112, opened 2023-03-19
Failing main tests / webonyx compatibility


Vincz:
| Q                | A
| ---------------- | -----
| Bug report?      | yes

Our tests fail with the latest version of `webonyx/graphql` because of the schema dumping test. 
Here: https://github.com/webonyx/graphql-php/blob/4ceb094b7850a586f04d487cc33702dd65ef9117/src/Utils/SchemaPrinter.php#L213

They are calling `getType()` for `Query`, `Mutation` and `Query`. The problem is that our type resolver raises an exception when a requested type is requested (here: https://github.com/overblog/GraphQLBundle/blob/f33382ac2bbdb750cbe448ac74cb1b6207e73c4f/src/Resolver/TypeResolver.php#L67).
I think that `webonyx/graphql` expect us to return `null` in this case (as shown here: https://github.com/webonyx/graphql-php/blob/4ceb094b7850a586f04d487cc33702dd65ef9117/src/Type/Schema.php#L305).

I'm not sure how to resolve this. Should we comply with webonyx requirement? If so, we should check everywhere we are using `getType` and assuming an raised exception. 

I don't have much time to look further, just wanted to write down my conclusion for further fix. 

@mcg-web Let me know what you think about this.

Vincz:
Fix in #1122

PR #1155, opened 2024-01-29
Fix profiler & update phpcs

Vincz:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | #... 
| License       | MIT

This fix the problem with the profiler if the schema doesn't have a subscription. 
It's related to https://github.com/overblog/GraphQLBundle/issues/1112 and the fact that Webonyx expects `null` from not found type with the `TypeResolver` while we raise an exception.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -116,7 +116,7 @@ jobs:
       - name: "Install PHP"
         uses: "shivammathur/setup-php@v2"
         with:
-          php-version: "8.2"
+          php-version: "8.1"
           coverage: "none"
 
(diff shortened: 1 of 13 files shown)


Vincz:
@mcg-web Hey mate! It would be great if you could check this (just a bunch of fixes and CS). Also, we should release a new version as the latests commits are fixing a bunch of problematics issues (and the SF 7 compat).

mcg-web (approved):
:+1:

Issue #1154, opened 2024-01-26
Example from doc not working under Symfony 6


hunhejj:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | no
| BC Break report? | no
| RFC?             | no
| Version/Branch   | master

I would like to call the Authorization Checker directly within my GraphQL type definition in YAML as suggested in [security/fields-public-control.md](https://github.com/overblog/GraphQLBundle/blob/a2235670f4b8207641a686da0177c5f92569a4bf/docs/security/fields-public-control.md?plain=1#L20). 

However this fails under Symfony 6 with the following message: 
> The "security.authorization_checker" service or alias has been removed or inlined when the container was compiled. You should either make it public, or stop using the container directly and use dependency injection instead.

Hence the doc should be updated. Is there any nice way to achieve the above besides of creating my own dummy wrapper around the `AuthorizationChecker`?

maximecolin:
Hi @hunhejj 

Indeed calling `security.authorization_checker` is no longer possible since Symfony remove or inline services. The documentation is outdated.

Instead you can use `@=hasRole(['ROLE_FOOBAR')` to check if the user has a particular role or `@=hasAnyRole(['ROLE_FOO', 'ROLE_BAR'])` to check if the user has one of these role.

You can find documentation about these expression here : https://github.com/overblog/GraphQLBundle/blob/master/docs/definitions/expression-language.md#hasrole

And if you want to check roles against an object, you can use [`@=hasPermission`](https://github.com/overblog/GraphQLBundle/blob/master/docs/definitions/expression-language.md#haspermission) or [`@=hasAnyPermission `](https://github.com/overblog/GraphQLBundle/blob/master/docs/definitions/expression-language.md#hasanypermission)

PR #1156, opened 2024-01-29
Do not inject the authorization checker service in GraphQL types

hunhejj:
In Symfony 6+ `security.authorization-checker` is not a public service anymore, hence the previous example won't work.

However the expression language supports using `isGranted` directly hence I propose to update the documentation.


hunhejj:
Addresses https://github.com/overblog/GraphQLBundle/issues/1154

Issue #1152, opened 2024-01-17
MetadataParser will throw exception if you have Symfony PHP Configurations in your bundle


cfv1000:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? |no
| BC Break report? |no
| RFC?             |no
| Version/Branch   | dev-master,1.0

I created the following boilerplate https://github.com/cfv1000/symfony-graphql/pull/1

Given a bundle with a PHP Configuration, the command  php bin/console graphql:validate will throw a MetadataParser exception at line 261

<img width="379" alt="image" src="https://github.com/overblog/GraphQLBundle/assets/3073353/80551498-7187-4a1e-98f2-7480c5db1bb2">

Under the following conditions
- Configure a bundle in symfony with working Overblog/GraphQL Bundle configuration
- create a .php configuration in the bundle

The library will break a feature supported by Symfony

GrozescuRares:
I am also interested into this topic.

Vincz:
Hi @cfv1000, I downloaded your projet and run it without error.  
Then I changed the `graphql.yaml` file into a `graphql.php` file like this: 
````php
<?php

namespace Symfony\Component\DependencyInjection\Loader\Configurator;

use function Symfony\Component\DependencyInjection\Loader\Configurator\param;

return static function (ContainerConfigurator $container): void {
    $container->extension('overblog_graphql', [
        // when using the param() function, you only have to pass the parameter name...
        'definitions' => [
            'schema' => [
                'query' => 'RootQuery'
            ],
            'mappings' => [
                'types' => [
                    ['type' => 'attribute', 'dir' => '%kernel.project_dir%/src/ApplicationBundle', 'suffix' => null]
                ]
            ]
        ]
    ]);
};
````

and run it again, and didn't get any error.  
Do you mind to update your project with the configuration that leads to the error?

cfv1000:
@Vincz  - there is an open pull request - if you change the branch from main to that pull request branch, and run the command again you should get the same error as i had

Vincz:
Ok, I understand. When we use a mapping with attributes, we expect the configured directories to contain only PHP files containing classes.  
That's why you get this error.  
One solution is to exclude this directories from the mapping parsing or be more specific on the directories containing your mapping.  
We could also check if the class does exist before trying to parse it in the `MetadataParser`. Feel free to submit a PR if you want, otherwise, I'll add this later.

PR #1157, opened 2024-01-30
fix issue with parsing invalid files

cfv1000:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   |no
| Fixed tickets | #1152 
| License       | MIT

- in the case where the reflection of a class fails, we should skip that file. I don't believe it would be the responsibility of the library to perform integrity checks on a php file
- updated tests - given 4 files, 3 which would not contain valid annotations / attributes on a class, the 4th file should still be processed.


src/Config/Parser/MetadataParser/MetadataParser.php
```diff
@@ -134,6 +134,8 @@ private static function processFile(SplFileInfo $file, ContainerBuilder $contain
             }
 
             return $preProcess ? self::$map->toArray() : $gqlTypes;
+        } catch (ReflectionException $e) {
```
Vincz, line 137: I'm not sure about this. What if there is other kinds of `ReflectionException` ? They would get silenced, wont they?
Are we 100% sure that this `catch` can only be trigger by invalid files?

cfv1000: reflection exception will happen where it fails to reflect using ReflectionClass. That will usually mean that the file is invalid or contains no class. 

So we have 2 situations
- when the file contains no class - this is the behaviour w want.
- if the file is invalid, the php processor will be responsible of throwing the exceptions. 

There could be a bigger feature - if we would want to validate the annotations / attributes at preParse stage. 

But i don't believe it's in the scope of the current PR

I tried an alternative with class_exists, but it will cause regression on other scenarios in the unit tests (i didn't take notes of the regressions)

Vincz (approved):

PR #1172, opened 2024-03-29
Cleanup Input default value and Fix #1171

### LINKED ISSUE #1171: Default field value for Input Object
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Version/Branch   | 1.1.0

The changes introduce in this PR https://github.com/overblog/GraphQLBundle/pull/1143, breaks the usage of default value in PHP for Input fields.   @deeky666 

```
#[GQL\Input]
class TestInput {
    #[GQL\Field(type: "String")]
    public string myField = "default_value"; // <-- This doesn't work anymore
}
```

Either the `MetadataParser` should check this when generating the type and add a `defaultValue` on the input field, or this field should be skip in the `ArgumentsTransformer`. 
Adding an `FieldInput` annotation/attribute should also be considered (as currently, the `#GQL\Field` does not allow a default value).

Vincz:
| Q             | A
| ------------- | ---
| Bug fix?      | kind of
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | #1171
| License       | MIT

- Handle default value on Input field and introduce a new attribute #[InputField]
- Update docs with example with attributes instead of annotation
- Rename `default` to `defaultValue` to be consistent in #[GQL\Arg] attribute (and deprecate the old one)

### DIFF
--- a/docs/annotations/annotations-reference.md
+++ b/docs/annotations/annotations-reference.md
@@ -591,13 +591,15 @@ class Hero {
 
 This annotation is used on _class_ to define a GraphQL interface.
 
-Required attributes:
-
--   **resolveType** : An expression to resolve the types
-
 Optional attributes:
 
+-   **resolveType** : An expression to resolve the types
 -   **name**  : The GraphQL name of the interface (default to the class name without namespace)
+
+If the `resolveType` attribute is not set, the service `overblog_graphql.interface_type_resolver` will be used to try to resolve the type automatically based on types implementing the interface and their associated class.  
+The system will register a map of interfaces with the list of types and their associated class name implementing the interface (the parameter is named `overblog_graphql_types.interfaces_map` in the container) and use it to resolve the type from the value (the first type where the class `instanceof` operator returns true will be used).  
+
+```php
 
 ## @Scalar
 
--- a/docs/annotations/annotations-reference.md
+++ b/docs/annotations/annotations-reference.md
@@ -599,8 +599,6 @@ Optional attributes:
 If the `resolveType` attribute is not set, the service `overblog_graphql.interface_type_resolver` will be used to try to resolve the type automatically based on types implementing the interface and their associated class.  
 The system will register a map of interfaces with the list of types and their associated class name implementing the interface (the parameter is named `overblog_graphql_types.interfaces_map` in the container) and use it to resolve the type from the value (the first type where the class `instanceof` operator returns true will be used).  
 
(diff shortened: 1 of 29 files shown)


docs/annotations/annotations-reference.md
```diff
@@ -578,13 +591,15 @@ class Hero {
...
 
 -   **resolveType** : An expression to resolve the types
+-   **name**  : The GraphQL name of the interface (default to the class name without namespace)
 
-Optional attributes:
+If the `resolveType` attribute is not set, the service `overblog_graphql.interface_type_resolver` will be used to try to resolve the type automatically based on types implementing the interface and their associated class.  
+The system will register a map of interfaces with the list of types and their associated class name implementing the interface (the parameter is named `overblog_graphql_types.interfaces_map` in the container) and use it to resolve the type from the value (the first type where the class `instanceof` operator returns true will be used).  
 
--   **name**  : The GraphQL name of the interface (default to the class name without namespace)
+```php
```
mcg-web, line 602:
```suggestion
```

src/Annotation/Arg.php
```diff
@@ -37,19 +37,33 @@ final class Arg extends Annotation
...
+    public function __construct(string $name, string $type, ?string $description = null, $defaultValue = null, $default = null)
     {
         $this->name = $name;
         $this->description = $description;
         $this->type = $type;
+        $this->defaultValue = $defaultValue;
         $this->default = $default;
+
+        if ($this->defaultValue === null && $this->default !== null) {
+            trigger_deprecation('overblog/graphql-bundle', '1.3', 'The "default" attribute on @GQL\Arg or #GQL\Arg is deprecated, use "defaultValue" instead.');
```
mcg-web, line 65: Use
```suggestion
@trigger_error(sprintf('....'), \E_USER_DEPRECATED);
```
instead of `trigger_deprecation` since it require [deprecation-contracts](https://github.com/symfony/deprecation-contracts)

docs/annotations/annotations-reference.md
Vincz: Fixed!

src/Annotation/Arg.php
Vincz: Fixed!

mcg-web (approved):
Thank you @Vincz !

Issue #1180, opened 2024-04-15
Help: Unclear how to use validation "link" with mutation/input objects


joesaunderson:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | no
| BC Break report? | no
| RFC?             | no
| Version/Branch   | 1

I am trying to use the validation linking as documented [here](https://github.com/overblog/GraphQLBundle/blob/master/docs/validation/index.md#linking-to-class-constraints) but I am struggling, and getting exceptions. 

Mutation:

```
EntityMutation:
    type: object
    heirs:
        - Mutation
    config:
        fields:
            edit_Entity:
                type: "Entity!"
                args:
                    entityId:
                        type: "ID!"
                    set:
                        type: "EntityEditInput!"
                        validation: cascade
                resolve: '@=query("App/Resolver::editEntity", args)'
```

Input Object

```
EntityEditInput:
    type: input-object
    config:
        fields:
            displayName:
                type: "String"
                validation:
                    link: 'App\Entity::$Name'
```

Exception

```
Property Overblog\\GraphQLBundle\\Validator\\ValidationNode::$Name does not exist
```

I noted this was mentioned [here](https://github.com/overblog/GraphQLBundle/blob/master/docs/validation/index.md#context-of-linked-constraints) but can you help me understand what the "correct" way of linking class constraints here would be (using an input object).

joesaunderson:
@Vincz @mcg-web, sorry for the direct tag. Can you tell me if I'm doing something obviously wrong here?

Ultimately, we wan't to use existing rules to validate fields within an input object, which is then used in a mutation.

joesaunderson:
Further debugging shows this is a naming / case sensitivity thing... the property on our entity was named `$Name` where the input property was called `displayName`

It wasn't obvious to me they needed to match identically - is it worth calling that out in the docs? Or even better, breaking the dependency between the input field name and the entity field name being identical?

PR #1181, opened 2024-04-15
Allow for linked validations to be named differently to property

joesaunderson:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes/no
| Documented?   | yes/no
| Fixed tickets | https://github.com/overblog/GraphQLBundle/issues/1180
| License       | MIT

We were overriding the `property` variable with the `property` from the `link`... this meant that the property (of an input or argument) had to match the property of the linked validation too.

This fixes that, allowing them to be separate, meaning you can have a property named `foo` and link it to a validation named `bar`.

Not sure if this is a breaking change... as technically the bug prevented people from having different properties -> entity properties, so there shouldn't be any skew anyway.

### DIFF
--- a/src/Validator/InputValidator.php
+++ b/src/Validator/InputValidator.php
@@ -169,7 +169,7 @@ final class InputValidator
             foreach ($config as $key => $value) {
                 switch ($key) {
                     case 'link':
-                        [$fqcn, $property, $type] = $value;
+                        [$fqcn, $classProperty, $type] = $value;
 
                         if (!in_array($fqcn, $this->cachedMetadata)) {
                             /** @phpstan-ignore-next-line */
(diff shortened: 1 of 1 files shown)


Vincz:
Hi @joesaunderson! Good catch. Don't think is a BC either. Anyway, that was clearly not intended.
Thanks!

joesaunderson:
> Hi @joesaunderson! Good catch. Don't think is a BC either. Anyway, that was clearly not intended.
> 
> Thanks!

Thanks @Vincz. Do you know if this issue is intended too? Happy to raise another PR to fix if not. 

https://github.com/overblog/GraphQLBundle/issues/1182

Issue #1182, opened 2024-04-16
Validation is called on optional fields that are not passed


joesaunderson:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no
| Version/Branch   | 1

When validating a field like so:

```
    ...
    config:
        description: "An input object for editing a merchant"
        fields:
            name:
                type: "String"
                validation:
                    link: 'App\Entity::$name'
            description:
                type: "String"
```

And using that in a mutation like

```
mutation {
  edit(set: {description: "Foo"}) {
    id
  }
}
```

I am getting an error for validation on the `name` field, which uses the Symfony "Assert/NotBlank" rule.

```
 "extensions": {
        "validation": {
          "set.name": [
            {
              "message": "Name must not be blank",
              "code": "c1051bb4-d103-4f74-8988-acbcafc7fdc3"
            }
          ]
        }
      },
```

Should validation be called for optional fields that are not passed in?

PR #1184, opened 2024-05-01
Don't validate fields that were not passed

joesaunderson:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes/no
| Documented?   | no
| Fixed tickets | https://github.com/overblog/GraphQLBundle/issues/1182
| License       | MIT

This fixes the validation rules so that when passing a partial input object, only the fields passed are validated.

### DIFF
--- a/src/Validator/InputValidator.php
+++ b/src/Validator/InputValidator.php
@@ -144,6 +144,7 @@ final class InputValidator
                 continue;
             }
             
+            
             $config = static::normalizeConfig($arg['validation'] ?? []);
 
             if (isset($config['cascade']) && isset($inputData[$property])) {
@@ -181,7 +182,6 @@ final class InputValidator
                             /** @phpstan-ignore-next-line */
                             $this->cachedMetadata[$fqcn] = $this->defaultValidator->getMetadataFor($fqcn);
                         }
-
 
                         // Get metadata from the property and it's getters
                         $propertyMetadata = $this->cachedMetadata[$fqcn]->getPropertyMetadata($classProperty);
--- a/src/Validator/InputValidator.php
+++ b/src/Validator/InputValidator.php
@@ -182,6 +182,7 @@ final class InputValidator
                             $this->cachedMetadata[$fqcn] = $this->defaultValidator->getMetadataFor($fqcn);
                         }
 
+
                         // Get metadata from the property and it's getters
                         $propertyMetadata = $this->cachedMetadata[$fqcn]->getPropertyMetadata($classProperty);
 
--- a/src/Validator/InputValidator.php
+++ b/src/Validator/InputValidator.php
@@ -138,14 +138,12 @@ final class InputValidator
 
         foreach ($fields as $name => $arg) {
             $property = $arg['name'] ?? $name;
+            $config = static::normalizeConfig($arg['validation'] ?? []);
 
             if (!array_key_exists($property, $inputData)) {
                 // This field was not provided in the inputData. Do not attempt to validate it.
                 continue;
             }
-            
-            
-            $config = static::normalizeConfig($arg['validation'] ?? []);
 
             if (isset($config['cascade']) && isset($inputData[$property])) {
                 $groups = $config['cascade'];
--- a/src/Validator/InputValidator.php
+++ b/src/Validator/InputValidator.php
@@ -138,6 +138,12 @@ final class InputValidator
 
         foreach ($fields as $name => $arg) {
             $property = $arg['name'] ?? $name;
+
+            if (!array_key_exists($property, $inputData)) {
+                // This field was not provided in the inputData. Do not attempt to validate it.
+                continue;
+            }
+            
             $config = static::normalizeConfig($arg['validation'] ?? []);
 
             if (isset($config['cascade']) && isset($inputData[$property])) {
--- a/tests/Functional/App/config/validator/mapping/Mutation.types.yml
+++ b/tests/Functional/App/config/validator/mapping/Mutation.types.yml
@@ -138,3 +138,11 @@ Mutation:
                         validation:
                             cascade:
                                 groups: ['group2']
+
+            onlyPassedFieldsValidation:
+                type: Boolean
+                resolve: '@=m("mutation_mock", args)'
+                args:
+                    person:
+                        validation: cascade
+                        type: Person!
--- a/tests/Functional/App/config/validator/mapping/Person.types.yml
+++ b/tests/Functional/App/config/validator/mapping/Person.types.yml
@@ -1,0 +1,14 @@
+Person:
+    type: input-object
+    config:
+        fields:
+            firstName:
+                type: String
+                validation:
+                    - NotBlank: ~
+                    - NotNull: ~
+            surname:
+                type: String
+                validation:
+                    - NotBlank: ~
+                    - NotNull: ~
--- a/tests/Functional/Validator/InputValidatorTest.php
+++ b/tests/Functional/Validator/InputValidatorTest.php
@@ -87,6 +87,22 @@ final class InputValidatorTest extends TestCase
         $this->assertTrue($result['data']['linkedConstraintsValidation']);
     }
 
+    public function testOnlyPassedFieldsValidated(): void
+    {
+        $query = '
+            mutation {
+                onlyPassedFieldsValidation(
+                    person: { firstName: "Joe" }
+                )
+            }
+        ';
+
+        $result = $this->executeGraphQLRequest($query);
+
+        $this->assertTrue(empty($result['errors']));
+        $this->assertTrue($result['data']['onlyPassedFieldsValidation']);
+    }
+
     public function testLinkedConstraintsValidationFails(): void
     {
         $query = '


Vincz:
Hi @joesaunderson!
Do you mind to add a bunch of tests please?

joesaunderson:
> Hi @joesaunderson! Do you mind to add a bunch of tests please?

I can try, there aren't many existing tests though, so not sure where I'll be starting from.

joesaunderson:
> > Hi @joesaunderson! Do you mind to add a bunch of tests please?
> 
> I can try, there aren't many existing tests though, so not sure where I'll be starting from.

Yep @Vincz this is a bit beyond my knowledge of this bundle, there's nothing to extend and I'm getting into the internals of the system.

Vincz (approved):

grossmannmartin:
Hi @Vincz, @joesaunderson 

I don't understand the need for this change.
Currently we're experiencing a changed and unexpected behavior in our application. 

I our use-case we're creating an order and filling customer data based on the input. 
Customer may choose a different delivery address. 

```
differentDeliveryAddress:
    type: "Boolean!"
deliveryFirstName:
    type: "String"
    validation:
        -   NotBlank:
                message: "Please enter first name of contact person"
                groups: "differentDeliveryAddressWithoutPreselected"
```

The validation group is then computed based on the value of the `differentDeliveryAddress` field - so when different delivery address is true (customer wishes to delivery goods to different address), the delivery fields become mandatory (regardless the input data). 

But after this change, the `deliveryFirstName` field is never validated if not passed in the mutation. This leads to the different behavior to data if passed and not, which is error-prone approach. 

Can you explain the use-case benefiting from this change? And do you have an idea how to solve the case I described?

joesaunderson:
The use case benefitting this change. Imagine a mutation like so:

```
edit_entity($entity)
```

Where `entity` is all the editable fields on that entity. 

If you only want to edit _some_ properties of the entity, you can pass only the properties you want to edit. 

Previously, if you did this, it ran the validation rules for every properly of that entity (or input object in this case) regardless of whether they were passed on not.

grossmannmartin:
But that way you have to count on the data you're passing down the application. So it's really easy to make a mistake and set data that should not pass the validation.

joesaunderson:
Not sure I fully understand your use case so forgive me... 

But isn't it correct that when not passing the `deliveryFirstName`, it shouldn't be validated? 

It should only be validated when it's passed (which is when `differentDeliveryAddress` is true?).

joesaunderson:
> But that way you have to count on the data you're passing down the application. So it's really easy to make a mistake and set data that should not pass the validation. 

Where does the counting element come in to it? This way the data you pass in is validated, the data you don't pass in isn't validated?

grossmannmartin:
If I set the differentDeliveryAddress = true, the I want to validate the deliveryFirstName regardless if passed in mutation. 
Because `differentDeliveryAddress ` means deliveryFirstName field has to be set.

Now I have to be sure I pass all field even though the schema do not require them, otherwise invalid data may be set

joesaunderson:
> If I set the differentDeliveryAddress = true, the I want to validate the deliveryFirstName regardless if passed in mutation. 
> 
> Because `differentDeliveryAddress ` means deliveryFirstName field has to be set.
> 
> 
> 
> Now I have to be sure I pass all field even though the schema do not require them, otherwise invalid data may be set

I understand it better now. Devils advocate... if you're essentially requiring that field when differentDeliveryAddress = true, shouldn't your application ensure it's always passed too? 

@Vincz suggested this behaviour could be a config setting, so maybe that could be an option? 

`validationMode: full | partial`

Vincz:
Hi @grossmannmartin, @joesaunderson 
Yes, in the end both of yours uses cases make sense. So, we should definitely add an option. 
What do you think, should be the default behavior?
I think we should be able to set the default behavior in the bundle config as suggested but we should also be able to override this in the input config itself. 
Does any of you want to take care of this PR?

joesaunderson:
I should have some time to look at it Tuesday. Can you point me in the direction of where config is used in other places?

grossmannmartin:
Hi, thanks for understanding and the dialog :) 

From my point of view, the original behavior is better as a default, because otherwise the rest of the application has to know what was passed as arguments and it's a behavior that already was.
But I am completely ok with the opposite as default as long as it's configurable :)

If necessary, I can make some time at the end of the week... sooner is not possible for me, sorry

joesaunderson:
@ITernovtsii see https://github.com/overblog/GraphQLBundle/issues/1185

I allowed the behaviour to be configurable, but I haven't had time to write tests / take it forward. 

Feel free to take it on if this is pressing

ITernovtsii:
@joesaunderson I think this change should be removed from the released version first, as "partial" mode will not work properly anyway, 
There is no reason to build the switch between 2 modes while only one mode works properly.
I reviewed the core a bit and don't see any simple solution for how this can be implemented without the validation break.
It looks like we may need metadata for each object, which will slow down request processing, and require significant changes in the code.

joesaunderson:
> @joesaunderson I think this change should be removed from the released version first, as "partial" mode will not work properly anyway, 
> 
> There is no reason to build the switch between 2 modes while only one mode works properly.
> 
> I reviewed the core a bit and don't see any simple solution for how this can be implemented without the validation break.
> 
> It looks like we may need metadata for each object, which will slow down request processing, and require significant changes in the code.

Partial mode does work, and is working for our use case in a production app for millions of users 😀

joesaunderson:
I think both your scenarios are when an input object is passed as an array, so maybe we need to adjust the check for that (does the field exist on any object in the array).

ITernovtsii:
Sorry, but I disagree that it makes sense to keep it because it works in some cases. Especially taking into account that bypass validation can be considered as a security issue.
Potentially, it can be easily solved on application level by ignoring null fields.

Issue is reproducible not only with arrays, but with any schema where same input type used twice in the request.

For example, if you have address input type, and passing it as person's deliveryAddress plus as homeAddress. _User who knows about this issue will be able to bypass validation rules, and set invalid county code to delivery address._
And those two types may exist in different depth levels of input data structure. So, it will require to iterate through the whole data tree to identify if there's object of the same type.

And I don't think it will be as simple as adjust this check, as it breaks consistency. Same object will be validated differently based on another object in the request.

joesaunderson:
There's was also another suggestion from @Vincz so allow the validation mode to be set at the config level (PR above) or at a mutation level.

Maybe you could extend my PR to add that too, allowing it to be set at a mutation by mutation basis?

joesaunderson:
> @joesaunderson 
> 
> @Vincz 
> 
> @grossmannmartin 
> 
> 
> 
> This change should not have been accepted. It's not fixing a bug; it's introducing one.
> 
> 
> 
> @joesaunderson in your [initial issue](https://github.com/overblog/GraphQLBundle/issues/1182), you showed this example:
> 
> 
> 
> ```yml
> 
> config:
> 
>     description: "An input object for editing a merchant"
> 
>     fields:
> 
>         name:
> 
>             type: "String"
> 
>             validation:
> 
>                 - NotBlank: ~
> 
>         description:
> 
>             type: "String"
> 
> ```
> 
> 
> 
> So here you are saying that `name` should not be a blank string, a blank array, `false`, or `null`. This means that `name` MUST be provided; otherwise, it's `null` and is invalid.  Then you pass this value, where `name` is missing:
> 
> 
> 
> ```graphql
> 
> mutation {
> 
>   edit(set: {description: "Foo"}) {
> 
>     id
> 
>   }
> 
> }
> 
> ```
> 
> And the validation fails, which is a perfectly correct behavior. Am I missing something? If you don't want to provide `name`, why make it mandatory? Because that's [precisely what `Assert/NotBlank` does](https://symfony.com/doc/current/reference/constraints/NotBlank.html#allownull:~:text=not%20equal%20to%20a%20blank%20string%2C%20a%20blank%20array%2C%20false%20or%20null).
> 
> 
> 
> 
> 
> What you should have done instead is this:
> 
> ```yml
> 
> config:
> 
>     description: "An input object for editing a merchant"
> 
>     fields:
> 
>         name:
> 
>             type: "String"
> 
>             validation:
> 
>                 - NotBlank:
> 
>                     allowNull: true # add this option
> 
>         description:
> 
>             type: "String"
> 
> ```
> 
> In human language, this means:
> 
> 1. If `name` is NOT provided, that's ok, because `null` is now allowed.
> 
> 2. If `name` is provided, then it should not be an empty string, an empty array or `false`.
> 
> 
> 
> This way, you can achieve your goal of not validating fields that are not provided.
> 
> 
> 
> I will open a roll-back PR because the current behavior is not correct and is counterintuitive. If we look into your test, we will see this:
> 
> ```yml
> 
> Person:
> 
>     type: input-object
> 
>     config:
> 
>         fields:
> 
>             firstName:
> 
>                 type: String
> 
>                 validation:
> 
>                     - NotBlank: ~
> 
>                     - NotNull: ~
> 
>             surname:
> 
>                 type: String
> 
>                 validation:
> 
>                     - NotBlank: ~
> 
>                     - NotNull: ~
> 
> ```
> 
> The validation constraints say the fields are required, yet we skip validation. This raises a logical question: why do we even need the `NotBlank` and `NotNull` constraints then?

I haven't worked on this for quite some time, but the thinking behind this was... if you want to mutate an entity that has say 40 properties

How do you mutate it without having to spread (pass in) every property, every time.

joesaunderson:
> @joesaunderson
> 
> 
> 
> > How do you mutate it without having to spread (pass in) every property, every time.
> 
> 
> 
> If you don't want to pass in every property, don't mark them **required**, simple as that. You configured your fields as `NotBlank`, which makes them required. Just remove that constraint.

I suppose it's how you interpret that rule. 

I read that as, if passed, not blank. Not, it has to be passed (required) every time. 

Otherwise, you need to fetch the entire object, in order to mutate a single field. 

Say you want to update just the name, with a simple form. You'd be over fetching 39 properties for the sake of updating one. 

But I can see both sides, hence the previous discussion of a 'mode' that could be configured.

PR #1185, opened 2024-05-14
Add the concept of "validation_mode"

joesaunderson:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| Fixed tickets | N/A
| License       | MIT

Adds the concept of `validation_mode` at the default config level.

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -80,7 +80,7 @@ final class Configuration implements ConfigurationInterface
 
     private function validationModeSection(): EnumNodeDefinition
     {
-        $builder = new TreeBuilder('validation_mode', "enum");
+        $builder = new TreeBuilder('validation_mode', 'enum');
 
         /** @var EnumNodeDefinition $node */
         $node = $builder->getRootNode();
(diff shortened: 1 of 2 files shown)


Vincz:
Excellent @joesaunderson! You are on the right track! Do you think it would be possible to also be able to manage the validation mode on the object itself (if it's too complicated, we'll try to implement in a future release).
If you can also add the test (with the same object applying both modes and validate the output).

TomasLudvik:
Hello @joesaunderson and @Vincz , is it possible to move his PR forward as it solves problem introduced in https://github.com/overblog/GraphQLBundle/pull/1184 ? We are locked on version 1.3.2 and are waiting for this to be fixed. Thank you.

joesaunderson:
I won't have the time I'm afraid, but someone can take this base and make the changes @Vincz suggests.

murtukov:
Closing this PR because it was opened based on the incorrect assumption that the validator asserts optional fields. In reality, if an optional field is omitted, it isn’t validated unless it’s explicitly marked with `Assert\NotBlank`.

PR #1190, opened 2024-07-04
Restrict to send JSON request as string (object expected)

vladdnepr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | 
| License       | MIT

Return value of getParsedBody must be array. But passed string will cause TypeError

Example

```
fetch("ENDPOINT", {
  "headers": {
    "content-type": "application/json",
  },
  "body": "111",
  "method": "POST"
});
```

### DIFF
--- a/src/Request/Parser.php
+++ b/src/Request/Parser.php
@@ -54,7 +54,7 @@ final class Parser implements ParserInterface
 
                 $parsedBody = json_decode($body, true);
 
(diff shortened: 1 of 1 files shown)


mcg-web (changes requested):
Thank you @vladdnepr, can you add a test covering this case please ?

PR #1203, opened 2025-05-27
add kernel.reset tag to allow clearing memory between request

Nyholm:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | 
| License       | MIT

When you run your application in an environment like FrankenPHP, RoadRunner etc, then you need to clear memory from one request to another. Using tag `kernel.reset` is the best way to tell Symfony to run the `reset()` method between requests.

### DIFF
--- a/src/Request/Executor.php
+++ b/src/Request/Executor.php
@@ -196,4 +196,9 @@ class Executor
             Events::POST_EXECUTOR
         )->getResult();
     }
+
+    public function reset(): void
+    {
+        $this->schemas = [];
+    }
 }
--- a/src/Resolver/TypeResolver.php
+++ b/src/Resolver/TypeResolver.php
@@ -76,6 +76,11 @@ class TypeResolver extends AbstractResolver
         return $type;
     }
 
+    public function reset(): void
+    {
+        $this->cache = [];
+    }
+
     protected function supportedSolutionClass(): ?string
     {
         return Type::class;
--- a/src/Resources/config/services.yaml
+++ b/src/Resources/config/services.yaml
@@ -21,6 +21,8 @@ services:
         calls:
             - ["setMaxQueryComplexity", ["%overblog_graphql.query_max_complexity%"]]
             - ["setMaxQueryDepth", ["%overblog_graphql.query_max_depth%"]]
+        tags:
+            - { name: kernel.reset, method: reset }
 
     Overblog\GraphQLBundle\Definition\Builder\SchemaBuilder:
         arguments:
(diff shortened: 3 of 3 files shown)


mcg-web (approved):
Thank you @Nyholm ! Sorry for the late reply.

MGDSoft:
When running the latest update, I get the error I mentioned. Executed with symfonyruntime & runtime/swoole

image

armetiz:
Since updating my project to this commit tag, the tests are failing on GitHub CI, but I can't reproduce the issue locally...

I spent two hours trying to figure out why there's a reset from GitHub CI when there isn't one from my local machine. Without success.

The error is : "RuntimeException: At least one schema should be declared.".

If you guys have a clue, it will be appreciate.

Note : I think the problem is coming from our tests implementations.

Nyholm:
Are you using Symfony KernelTestCase (or WebTestCase) and try to do more that one request per test case?

VrbovskyM:
After updating the package to version with this change, we experience error `"At least one schema should be declared."` when running local tests. We do call graphql request more than once in some tests.

Nyholm:
When you are instantiating your services you get a `schemaBuilder`. That is turned into a schema when you execute it. 

When the kernel is reset, we clear up the memory (as we should). When you run your test, you reuse the same service (now without a schema builder). This does not happen in prod because then you destroy your objects properly. 

How to solve it? 

- Either make sure you instantiate your services again (just like prod) for your second request
- Do not make a second request. 

There is maybe a better way to solve it, if you find it, please update this ticket.

OlehNechyporuk:
Hi 👋 thanks for the bundle and recent changes around kernel.reset.

We are running GraphQLBundle with RoadRunner and attribute-based schema definitions. After the change where overblog_graphql.request_executor is tagged with kernel.reset, we consistently hit this issue:

first request works

on second request executor throws
“At least one schema should be declared.”

Root cause seems to be:

Executor::reset() clears internal $schemas

in long-running workers (RoadRunner / FrankenPHP) the container is not rebuilt

schema builders from attribute mappings are not re-applied after reset

so executor stays empty after first request

This makes attribute-based schemas unusable in long-running runtimes unless we override executor behavior.

Question: what is the recommended way to handle this case with attribute mappings + RoadRunner?
Should schema builders be re-registered automatically after reset, or should reset be configurable/optional for executor?

Thanks 🙏

eexit:
Our Behat stack is probably not well isolated and all, but since version 1.9.0, we faced this issue as well. Until we improve our Behat tests stack, I had to add a env test only compiler pass:

```php
final class GraphQLNonSharedExecutorPass extends CompilerPass
{
    public function process(ContainerBuilder $container): void
    {
        if (!$container->hasDefinition(Executor::class)) {
            return;
        }

        $definition = $container->getDefinition(Executor::class);
        $definition->setShared(false);
    }
}
```

I'm not sure this is the best approach, but if this can help 🤷🏻‍♂️

Issue #1168, opened 2024-03-25
Multi level Input-object validation issue


morkers:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no
| Version/Branch   | 0.13.7

There seems to be an issue validating array of input-objects.

Having a simple mutation:
```
Mutation:
    type: object
    config:
        fields:
            testMutation:
                type: Boolean
                resolve: "@=mutation('testMutation', [args])"
                args:
                    fields:
                        type: '[TestInput]'
                        validation: cascade
TestInput:
    type: input-object
    config:
        fields:
            subField:
                type: SubTestInput
                validation: cascade

SubTestInput:
    type: input-object
    config:
        fields:
            price:
                type: Float
                validation:
                    -  Positive: ~
```
Then sending mutation i don't get the error i expect. The first `SubTestInput` is not validated.
```
mutation{
    testMutation(fields: [
        {
            subField: {
                price: -1
            }
        },
        {
        }
    ])
}
```
As the second/last `TestInput` does not have `SubTestInput` defined, validation metadata is overriden with metadata that doesn't have any constraints registered and validation for first `SubTestInput` is not done.

Vincz:
I'm not familiar with the validator component. Maybe @murtukov can look into it?

kedramon:
same for me, none of child validations run for array of input-object

jsakars:
No updates on this?

murtukov:
@Vincz @jsakars I will take a look at it today

murtukov:
@jsakars @morkers a quick update: I investigated the issue and found the bug. I'll try to provide a fix in the following days

PR #1215, opened 2025-11-02
Fix `input-object` validation issue

murtukov:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | #1168 
| License       | MIT

As mentioned in the ticket, there was an issue with the validation metadata for collections of input objects: only the metadata of the last item in the collection was used to validate all items in the collection.

This change introduces metadata caching, so that the same object is used for all elements in the collection.

### DIFF
--- a/tests/Functional/Validator/ExpectedErrors.php
+++ b/tests/Functional/Validator/ExpectedErrors.php
@@ -88,23 +88,23 @@ final class ExpectedErrors
                 'addresses[0].country.officialLanguage' => [
                     0 => [
                         'message' => 'The value you selected is not a valid choice.',
-                        'code' => '8e179f1b-97aa-4560-a02f-2a8b42e49df7'
+                        'code' => '8e179f1b-97aa-4560-a02f-2a8b42e49df7',
                     ],
                 ],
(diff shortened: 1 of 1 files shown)


src/Validator/InputValidator.php
```diff
@@ -154,18 +167,29 @@ private function buildValidationTree(ValidationNode $rootObject, iterable $field
...
                 if (!empty($groups)) {
                     $valid->groups = $groups;
                 }
 
+                // Apply the Assert/Valid constraint for a recursive validation.
+                // For more details see https://symfony.com/doc/current/reference/constraints/Valid.html
                 $metadata->addPropertyConstraint($property, $valid);
+
+                // Skip the rest as the validation was delegated to the nested object.
+                continue;
```
murtukov: There is no need to proceed and add any more constraints if it was set to `cascade`

murtukov:
@mcg-web, since you already approved the same PR for the branch 0.13, I will consider this one approved as well.

PR #1237, opened 2026-01-28
Test on PHP 8.5

### LINKED ISSUE #1221: PHP 8.5 Compatibility Error: Attribute applied to abstract class.
In PHP 8.5 and newer, it is no longer permitted to use an attribute on an abstract class. Currently this is being done in _at least_ the following location:

https://github.com/overblog/GraphQLBundle/blob/d9c6f53ea923c92c14cbfa20905cf36f6656d003/src/Annotation/Builder.php#L17

This prevents the package from working on PHP 8.5 completely, which is a problem for platforms which only support the current PHP version.

Nyholm:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Documented?   | no
| Fixed tickets | fix #1221
| License       | MIT

Lets see if we can bring PHP 8.5 support. 

The only thing we need to do is remove the attribute from the abstract class. Absolute worst case scenario is that some extended this annotation and did not add an attribute themselves. 

Looking at [this issue](https://github.com/php/php-src/issues/19694) from PHP src, it is save to remove because "it was not really working" in the first place.


martiis:
Looks like this is the way <3

Nyholm:
Thank you

PR #1240, opened 2026-03-11
Fix isAnonymous() always returning false on Symfony 6.x

### LINKED ISSUE #1205: Access denied on this field on query built with PHP way + attributes
Hi,

I'm trying to migrate from a YAML implementation to a full PHP implementation.
I well know the solution by using this bundle for years now.
But, with this config, new to me, I'm facing an issue accessing the fields on root query/mutation.

Here's my setup using the 1.7.0.

Image
Image
Image
Image

The graphql:debug does not show the query, but with GraphiQL, it seems to be registered:
Image

And doing this query throws an access error:
Image

I tried to so the same thing on a mutation. Same result.
IsPublic is not supported on providers or queries I guess.

So, at this point, I have no more ideas to find a solution.

Thank you a lot in advance for helping!

mfaivre-simplis:
## Summary

Fixes #1205

`BaseSecurity::isAnonymous()` calls `isGranted('IS_AUTHENTICATED_ANONYMOUSLY')`, which was removed from `AuthenticatedVoter` in Symfony 6.x (deprecated since 5.4) : https://github.com/symfony/symfony/blob/6.0/UPGRADE-6.0.md#security:~:text=Remove%20AuthenticatedVoter%3A%3AIS_AUTHENTICATED_ANONYMOUSLY%20and%20AuthenticatedVoter%3A%3AIS_ANONYMOUS%2C%20use%20AuthenticatedVoter%3A%3APUBLIC_ACCESS%20instead.

The access decision manager does not work and returns `false`. Every field annotated with `#[GQL\Access('isAnonymous()')]` gets "Access denied to this field" for all requests.

The fix uses `PUBLIC_ACCESS` on Symfony 6.x, which has the same behavior (`IS_AUTHENTICATED_ANONYMOUSLY` and `PUBLIC_ACCESS` both always grant access) and is the [official replacement](https://symfony.com/doc/6.4/security.html#checking-to-see-if-a-user-is-logged-in-is-authenticated-fully).

## Changes

- `src/Security/Security.php`: use `PUBLIC_ACCESS` on Symfony >= 6.0, keep `IS_AUTHENTICATED_ANONYMOUSLY` for 5.x

## Reproduction

A minimal reproduction repo is available at: https://github.com/mfaivre-simplis/graphqlbundle-issue-1205

```php
#[GQL\Provider]
#[GQL\Access('isAnonymous()')]
final readonly class ConnectedUserQuery
{
    #[GQL\Query(name: 'connectedUser', type: 'Boolean!')]
    #[GQL\Access('isAnonymous()')]
    public function __invoke(): bool
    {
        return true;
    }
}
```

With Symfony 6.x, querying `{ connectedUser }` always returns `"Access denied to this field"` before this fix.

### DIFF
--- a/src/Security/Security.php
+++ b/src/Security/Security.php
@@ -104,7 +104,9 @@ abstract class BaseSecurity
 
     public function isAnonymous(): bool
     {
-        return $this->isGranted('IS_AUTHENTICATED_ANONYMOUSLY');
+        // IS_AUTHENTICATED_ANONYMOUSLY was removed from AuthenticatedVoter in Symfony 6.x.
+        // PUBLIC_ACCESS (added in Symfony 5.4) is the correct replacement and always grants access.
+        return $this->isGranted(Kernel::VERSION_ID >= 60000 ? 'PUBLIC_ACCESS' : 'IS_AUTHENTICATED_ANONYMOUSLY');
     }
 
     public function isAuthenticated(): bool
--- a/src/Security/Security.php
+++ b/src/Security/Security.php
@@ -104,9 +104,7 @@ abstract class BaseSecurity
 
     public function isAnonymous(): bool
     {
-        // IS_AUTHENTICATED_ANONYMOUSLY was removed from AuthenticatedVoter in Symfony 6.x.
-        // PUBLIC_ACCESS (added in Symfony 5.4) is the correct replacement and always grants access.
-        return $this->isGranted(Kernel::VERSION_ID >= 60000 ? 'PUBLIC_ACCESS' : 'IS_AUTHENTICATED_ANONYMOUSLY');
+        return $this->isGranted('PUBLIC_ACCESS');
     }
 
     public function isAuthenticated(): bool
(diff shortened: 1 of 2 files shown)


src/Security/Security.php
```diff
@@ -104,7 +104,9 @@ public function hasRole(string $role): bool
 
     public function isAnonymous(): bool
     {
-        return $this->isGranted('IS_AUTHENTICATED_ANONYMOUSLY');
+        // IS_AUTHENTICATED_ANONYMOUSLY was removed from AuthenticatedVoter in Symfony 6.x.
+        // PUBLIC_ACCESS (added in Symfony 5.4) is the correct replacement and always grants access.
+        return $this->isGranted(Kernel::VERSION_ID >= 60000 ? 'PUBLIC_ACCESS' : 'IS_AUTHENTICATED_ANONYMOUSLY');
```
Nyholm: Let's keep it simple. We dont support 5.3 or lower. 


```suggestion
        return $this->isGranted('PUBLIC_ACCESS');
```

Nyholm (commented):
Great. Thank you. But let's drop `IS_AUTHENTICATED_ANONYMOUSLY` completely.

src/Security/Security.php
mfaivre-simplis: Sure, it's done.

Nyholm (approved):

PR #1241, opened 2026-03-17
resettable schemas, getCurrentSchemaName for TypeResolver, and IsPublic support for InputObject fields

vlydev:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | yes
| License       | MIT

---                                                                                                                                                                             
  Summary         
                                                                                                                                                                                  
  This PR adds three independent improvements to the bundle, primarily motivated by long-running process compatibility (e.g. RoadRunner, Swoole, FrankenPHP) and better ACL
  control over input types.                                                                                                                                                       
   
  ---                                                                                                                                                                             
  1. Resettable schemas (SchemaBuilder, Executor, AbstractResolver)
                                                                                                                                                                                  
  What changed:
  - SchemaBuilder and Executor now implement ResetInterface                                                                                                                       
  - A new resettable: true option can be set per schema in config
  - ExtensibleSchema gained isResettable()/setIsResettable() methods                                                                                                              
  - AbstractResolver now properly separates factories ($solutionsFactory) from built instances ($solutions) and implements reset() — clearing only the built instances, not the   
  factories                                                                                                                                                                          
                                                                                                                                                                                  
  In long-running process environments (RoadRunner, Swoole, ReactPHP), the Symfony kernel's reset() cycle is called between requests. Previously, built schemas were cached via a 
  static variable inside the getBuilder() closure, which meant they survived resets entirely — causing stale type registries and memory growth over time.
                                                                                                                                                                                  
  This change moves schema caching into SchemaBuilder::$builders (an instance property), so reset() can selectively evict schemas marked as resettable: true. Non-resettable      
  schemas (the default) are kept across resets for performance. The same principle applies to AbstractResolver: factories are registered once at boot and survive resets, while
  resolved instances are cleared so they can be rebuilt fresh on the next request.                                                                                                
                  
```
  # config/packages/graphql.yaml
  overblog_graphql:                                                                                                                                                               
      definitions:                                                                                                                                                                
          schema:                                                                                                                                                                 
              default:                                                                                                                                                            
                  query: RootQuery                                                                                                                                                
                  resettable: true   # rebuilt on kernel reset         
```                                                                                                           
                                                                                                                                                                                  
  ---             
  2. TypeResolver::getCurrentSchemaName()                                                                                                                                         
                                                                                                                                                                                  
  What changed:
  - Added a public getter getCurrentSchemaName(): ?string alongside the existing setCurrentSchemaName()                                                                           
                                                                                                                                                                                  
  The current schema name is set internally during schema build but was previously inaccessible from outside the resolver. Services and extensions that need to apply             
  schema-specific logic (e.g. different ACL rules per schema, custom type loaders, multi-schema middleware) had no way to read which schema is currently being resolved without
  injecting additional state. This getter exposes what was already tracked internally, with zero behavioral change.                                                               
                  
  ---
  3. #[IsPublic] support on InputObject fields

  What changed:
  - InputObjectTypeDefinition now accepts a public key on field-level config
  - MetadataParser reads #[IsPublic] attributes (and @IsPublic annotations) on InputObject class properties and maps them to public in the field configuration                    
  - Using #[IsPublic] on a property without #[Field] throws InvalidArgumentException (consistent with the existing behavior for #[Access] on object types)                                                                                                                                            
                                                                                                                                                                                  
  #[IsPublic] was already supported on ObjectType fields and at the provider level, but InputObject fields had no equivalent. This gap meant that input fields carrying sensitive 
  data (e.g. an adminNotes field on a search input) could not be conditionally hidden via expression language — the only option was to create separate input types. This change
  brings InputObject field visibility in line with ObjectType field visibility, using the same expression-language mechanism already present in the bundle.

### DIFF
--- a/docs/definitions/schema.md
+++ b/docs/definitions/schema.md
@@ -74,7 +74,29 @@ overblog_graphql:
             # by graphql-php during static schema analysis.
             # These types names should be explicitly declare here
             types: []
+            # when true, the built schema is discarded whenever the
+            # container is reset (e.g. between requests in long-running
+            # workers like Symfony Runtime, RoadRunner or FrankenPHP
+            # worker mode). Useful when the schema depends on request
+            # scoped state. Defaults to false.
+            resettable: false
 ```
+
+## Resettable schemas
+
+By default, schemas are built once per worker and reused across requests. When running Symfony in a long-running process (Symfony Runtime, RoadRunner, FrankenPHP worker mode, Swoole, etc.) and the schema depends on request scoped state, set `resettable: true` so the schema is rebuilt after each `kernel.reset`:
+
+```yaml
+overblog_graphql:
+    definitions:
+        schema:
+            default:
+                query: Query
+                mutation: Mutation
+                resettable: true
+```
+
+Non-resettable schemas (the default) are kept across resets, which preserves the build cost between requests.
 
 ## Batching
 
@@ -93,6 +115,7 @@ overblog_graphql:
             bar:
                 query: barQuery
                 mutation: barMutation
+                resettable: true
 ```
 
 **foo** schema endpoint can be access:
--- a/docs/security/fields-public-control.md
+++ b/docs/security/fields-public-control.md
@@ -66,3 +66,40 @@ AnObject:
 Have you noticed `typeName` and `fieldName` here? These variables are always set to the current
 type name and current field name, meaning you can apply a per field `public` setting on all the
 fields with one line of yaml.
+
+## Input object fields
+
+`public` is also supported on `input-object` fields. When the expression resolves to `false`,
+the field is removed from the input type — it is hidden from introspection and rejected if a
+client tries to submit it.
+
+```yaml
+HeroInput:
+    type: input-object
+    config:
+        fields:
+            name:
+                type: "String!"
+            internalNote:
+                type: "String"
+                public: "@=isGranted('ROLE_ADMIN')"
+```
+
+With attributes, combine `#[GQL\Field]` with `#[GQL\IsPublic]` on the property:
+
+```php
+<?php
+
+use Overblog\GraphQLBundle\Annotation as GQL;
+
+#[GQL\Input]
+class HeroInput
+{
+    #[GQL\Field(type: "String!")]
+    public string $name;
+
+    #[GQL\Field(type: "String")]
+    #[GQL\IsPublic("isGranted('ROLE_ADMIN')")]
+    public ?string $internalNote = null;
+}
+```
--- a/src/Config/InputObjectTypeDefinition.php
+++ b/src/Config/InputObjectTypeDefinition.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Overblog\GraphQLBundle\Config;
 
 use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
+use Symfony\Component\Config\Definition\Builder\VariableNodeDefinition;
 
 use function is_string;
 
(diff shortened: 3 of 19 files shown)


Nyholm:
Oh, This looks awesome.

Thank you. I've asked some friends to give in a thorough review.

martiis:
Hi, thank you for MR, looking good! The only thing still missing is documentation for the new resettable config and update about `public` attribute

Nyholm:
@vlydev Can you please add some documentation and I'll be happy to merge

vlydev:
@Nyholm done, please check

vlydev:
@Nyholm check please and merge

MGDSoft:
Some AI agent is available to check this solution? 😭🤣🤣

vlydev:
Any updates?

vlydev:
@Nyholm @martiis ping

vlydev:
> Some AI agent is available to check this solution?

@MGDSoft 

The PR has three independent changes. None of them breaks backward compatibility, and every new behavior is opt-in or additive.

  1. Resettable schemas (for long-running workers: RoadRunner, FrankenPHP, Swoole)

  - Problem: the built schema was cached in a static $schema variable inside a closure. Nothing could clear it, so a stale schema stayed in the worker across all requests.
  - Change:
    - SchemaBuilder now caches built schemas in an instance property $builders and implements ResetInterface. reset() drops only schemas marked resettable.
    - New per-schema config option resettable: true|false, default false.
    - ExtensibleSchema gets isResettable() / setIsResettable().
    - Executor keeps builder closures ($schemaBuilders) separate from built schemas ($schemas). On reset it keeps only non-resettable ExtensibleSchema instances.
    - AbstractResolver keeps factories ($solutionsFactory) separate from built instances ($solutions). reset() clears the instances, and the factories stay registered.
      TypeResolver is tagged kernel.reset.
  - Default behavior (resettable: false) is unchanged. The schema is built once per worker.

  2. TypeResolver::getCurrentSchemaName()

  - Adds a public getter next to the existing setter. It returns the name of the schema currently being built or resolved.
  - Use case: per-schema logic (ACL, type loaders) without extra state. Behavior is unchanged.

  3. #[IsPublic] / public: on input-object fields

  - InputObjectTypeDefinition accepts public on fields, the same way object-type fields already do.
  - MetadataParser maps #[GQL\IsPublic] / @GQL\IsPublic on an #[GQL\Input] property to public in the field config.
  - #[IsPublic] without #[Field] throws InvalidArgumentException. This matches the existing behavior for object types.
  - At runtime the existing PublicFieldsFilterConfigProcessor removes the field when the expression is false. The field disappears from introspection and is rejected if a client
    sends it.

  How to verify

  1. Unit tests:
  git fetch origin pull/1241/head:pr-1241 && git checkout pr-1241
  composer install
  vendor/bin/phpunit --filter 'SchemaBuilderTest|ExtensibleSchemaTest|ExecutorTest|TypeResolverTest|TestMetadataParser'
  vendor/bin/phpunit   # full suite, nothing else should change
  2. Resettable schema, manual check in a sample app:
     - Set resettable: true on a schema. Call $executor->getSchema('default') twice with $container->get('services_resetter')->reset() in between. You should get different
       instances.
     - Repeat with resettable: false (or no option). You should get the same instance.
  3. getCurrentSchemaName(): inside any service that has TypeResolver injected, call it during a request to the admin schema. It should return 'admin'.
  4. Input field visibility:
     - Add internalNote: { type: String, public: "@=isGranted('ROLE_ADMIN')" } to an input type.
     - As an anonymous user: introspection does not list internalNote, and a query that sends it fails with "Field is not defined by type".
     - As an admin: the field is visible and accepted.
  5. Backward compatibility: existing configs without resettable and without input public should produce the same generated types (graphql:compile output diff is empty).

Issue #1193, opened 2024-09-03
ArgumentsValidationException logged as critical


nicolas-codemate:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | no
| Version/Branch   | 1.5.0

Hi,

I use the `ArgumentsValidationException` in mutations frequently to ensure my model adheres to business logic using Symfony asserts.

It's very handy because validation errors are properly displayed without much effort.

The problem is these kinds of exception are logged as critical (see https://github.com/overblog/GraphQLBundle/blob/a8c5661ece3acabf9535034fdf59bceed08b2567/src/EventListener/ErrorLoggerListener.php#L28)

In my opinion, it should behave exactly like `UserError` (logged as Error).

What do you think ?

Proposals :
- Make ArgumentsValidationException extends UserError 
- since ArgumentsValidationException implement ClientAware and marked as "ClientSafe"
```php
// Overblog\GraphQLBundle\EventListener\ErrorLoggerListener

    public function onErrorFormatting(ErrorFormattingEvent $event): void
    {
        // ...

        if ($exception instanceof UserWarning) {
            if ($exception->getPrevious()) {
                $this->log($exception->getPrevious(), LogLevel::WARNING);
            }

            return;
        }

        if ($exception instanceof ClientSafe) {
            if ($exception->getPrevious() && $exception->isClientSafe()) {
                $this->log($exception->getPrevious());
            }

            return;
        }

        $this->log($exception, LogLevel::CRITICAL);

```
- something else ?

murtukov:
Hi @nicolas-codemate, this is a nice proposal indeed. I will try to work on it over the following days.

PR #1248, opened 2026-08-10
fix: avoid critical logs for argument validation failures (#1193)

ousamabenyounes:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documented?   | no
| Fixed tickets | #1193
| License       | MIT

### Problem

`ArgumentsValidationException` represents invalid client input, but
`ErrorLoggerListener` treated it like an unknown server exception and logged it at
`CRITICAL`.

### Fix

Treat `ArgumentsValidationException` like `UserError`: log its previous cause at
`ERROR` when one exists, and do not emit a `CRITICAL` entry for a routine validation
failure without a previous cause.

The condition is deliberately limited to `ArgumentsValidationException`.
Other exceptions implementing `ClientAware` retain their existing `CRITICAL` logging,
because client-safe messages do not necessarily indicate client-caused failures.

### Test verification (RED → GREEN)

The two argument-validation cases fail on the unmodified base because they reach the
`CRITICAL` fallback:

```
FAILURES!
Tests: 11, Assertions: 10, Failures: 2, Risky: 1.
```

An additional regression case failed against the earlier generic `ClientAware`
implementation because the expected `CRITICAL` log was suppressed:

```
FAILURES!
Tests: 11, Assertions: 11, Failures: 1.
Method was expected to be called 1 time, actually called 0 times.
```

With the final scoped implementation:

```
OK (11 tests, 11 assertions)
```

The full suite remains iso-baseline: the same five pre-existing
`GraphDumpSchemaCommandTest` failures occur on `master`; the final run reports
`Tests: 715, Assertions: 1722, Failures: 5, Skipped: 8`.

### DIFF
--- a/src/EventListener/ErrorLoggerListener.php
+++ b/src/EventListener/ErrorLoggerListener.php
@@ -7,6 +7,7 @@ namespace Overblog\GraphQLBundle\EventListener;
 use GraphQL\Error\UserError;
 use Overblog\GraphQLBundle\Error\UserWarning;
 use Overblog\GraphQLBundle\Event\ErrorFormattingEvent;
+use Overblog\GraphQLBundle\Validator\Exception\ArgumentsValidationException;
 use Psr\Log\LoggerInterface;
 use Psr\Log\LogLevel;
 use Psr\Log\NullLogger;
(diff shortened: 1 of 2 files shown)


ousamabenyounes:
Updated in `9bbc592e`: the fix now applies only to `ArgumentsValidationException`; other client-safe exceptions retain their existing `CRITICAL` logging.

